PR #49, opened 2019-10-29
FIX: only run bids if not recon-only *and* not skip bids

WillForan:
Skip BIDS validation when either `--skip-bids-validation` or `--recon-only` are supplied. 
Previously, both were need both to skip validation.

N/A unless previous behavior is intended, then maybe a note in `--skip_bids_validation`'s parser help text


mattcieslak:
I ended up adding this fix to another patch, so we can close this PR. Could you send your info for the zenodo.json file? https://github.com/PennBBL/qsiprep/blob/master/.zenodo.json#L13 I can add it before the next release

WillForan:
Sorry it took so long!

Issue #75, opened 2019-11-20
CUDA driver/runtime version mismatch


eds-slim:
Hi,
I'm currently exploring `qsiprep v0.6.4` on Ubuntu 18.04 and encountered a problem with CUDA. Specifically, very early on, the pipeline throws the error
```
191120-19:50:44,55 nipype.workflow WARNING:
	 [Node] Error on "qsiprep_wf.single_subject_00012_wf.dwi_preproc_acq_NDDFC_run_01_wf.hmc_sdc_wf.eddy" (/work/qsiprep_wf/single_subject_00012_wf/dwi_preproc_acq_NDDFC_run_01_wf/hmc_sdc_wf/eddy)
191120-19:50:45,885 nipype.workflow ERROR:
	 Node eddy failed to run on host Ixion.
191120-19:50:45,886 nipype.workflow ERROR:
	 Saving crash info to /work/wdir/bids50/derivatives/qsiprep/sub-00012/log/20191120-194715_25077cc2-befc-4960-8775-7c4f7057509b/crash-20191120-195045-eckhard-eddy-67060859-878b-4c86-8521-bf03601ca462.txt
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 473, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 557, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 637, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 375, in run
    runtime = self._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/fsl/epi.py", line 766, in _run_interface
    runtime = super(Eddy, self)._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 758, in _run_interface
    self.raise_exception(runtime)
[...]

PR #84, opened 2019-12-03
[ENH] Upgrade to FSL 6

mattcieslak:
Based on #75 
 * Update Dockerfile to include FSL 6 instead of FSL 6
 * Use the CUDA ubuntu image as a base.

 - [x] Add a section on CUDA


mattcieslak:
This got added in #91

PR #136, opened 2020-06-17
[ENH] Enable HCP-style polarity-pair image averaging

mattcieslak:
The HCP pipelines run the same sampling scheme twice - once in each phase-encoding direction. Each phase-encoding direction version of the scheme is split into two files, resulting in 4 DWI scans. There are also a pair of EPI fieldmaps in the fmaps/ directory.

This PR adds the --distortion-group-merge option that can be 'none' (Default), 'average' (for HCP-style averaging) or 'concat' (produces a 4d series of the corrected input images).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.9.0beta1 (June 17, 2020)
+==========================
+
+ * Adds support for HCP lifespan sequences
+ * Introduces --distortion-group-merge option for combining paired scans
+
 0.8.0 (February 12, 2020)
 =========================
 
@@ -5,6 +11,7 @@
 
 0.7.2 (February 4, 2020)
 ========================
+
  * Fixed a bug in b=0 masking when images have high signal intensity in ventricles (#99)
 
 0.7.1 (January 29, 2020)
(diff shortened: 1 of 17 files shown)


cookpa:
Are the field maps required? The HCP Young Adult protocol has these, but the later protocols do not. The 2016 CCF protocol doesn't have any field maps before the dMRI

https://protocols.humanconnectome.org/CCF/

The HCP Aging and Development protocol acquires a SpinEchoFieldMap_[AP,PA] pair for topup, but I don't think they get gradient-echo field maps

https://www.humanconnectome.org/study/hcp-lifespan-aging/project-protocol/imaging-protocols-hcp-aging

mattcieslak:
If there are images in `fmaps/` that have the dwi scans in their `IntendedFor` then the images from `fmaps/` are used along with with b=0 images from the scans in `dwi/` in TOPUP. The Lifespan sequences I've been testing have AP/PA Spin Echo fieldmaps (3 b=0 images in each direction). These get combined with 3 b=0 images from each of the dwi scans. This results in 18 total images (3 from fmaps + 3 from run 1  + 3 from run 2) x 2 PE Directions being used in TOPUP, which takes a long time to run and should probably be changed.

If there are no fieldmaps in `fmaps/`, then 3 b=0 images are taken from each dwi scan and used in TOPUP. 

In all cases the b=0 images used are the first, middle and last b=0 images in the scan.

Issue #180, opened 2020-09-18
[BUG] eddy_cuda fails if ran with `--estimate_move_by_susceptibility`


fredrmag:
Hi,

This bug might be more related to FSL. But I will post it here anyway.

When running qsiprep v.0.11.0 using `eddy_cuda` with the flag `--estimate_move_by_susceptibility`. Our group get the following error:

![image (1)](https://user-images.githubusercontent.com/21278572/93598295-c4d9d700-f9bc-11ea-9d63-d087d7aab464.png)

fredrmag:
Might be related: https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=FSL;b671062b.1910

ethanknights:
Full log output is here in case useful:
[slurm.qsiprep.sub-CC110037.ses-01-gpu.log](https://github.com/PennBBL/qsiprep/files/5245818/slurm.qsiprep.sub-CC110037.ses-01-gpu.log)

mattcieslak:
Good find @fredrmag, can you confirm there's a NaN in the pixdim field in the header?

mattcieslak:
Are there NaNs in any of the inputs to eddy? If you're using fieldmaps there will likely be other image inputs. Have you had any successful runs with `estimate_move_by_susceptibility`? We're trying out this feature now too

fredrmag:
This participant do not have a fieldmap. Only other nifty input is the mask. Does not have NaN in the header. But, this NaN header should be fixed by Jesper in October 2019. So, if qsiprep is using the newest FSL, it should'nt be a problem?

fredrmag:
When eddy ran successfully with the `--estimate_move_by_susceptibility` flag, FSL 6.0.2 with eddy_cuda9.1 was used. We have not tested FSL 6.0.3 outside qsiprep.

fredrmag:
We created a singularity image of of the fix proposed by PR #181 using commit: 32462e9dc883012b53a868fe350fb31c948b2fac. @ethanknights tested this on a dataset, and we still get the same error message. I will attach the logs. 

![Untitled](https://user-images.githubusercontent.com/21278572/94913027-d1742a00-04a8-11eb-9f91-3e36554cc6ce.png)
[slurm.qsiprep.sub-CC110037.ses-01-gpu.log](https://github.com/PennBBL/qsiprep/files/5317270/slurm.qsiprep.sub-CC110037.ses-01-gpu.log)
[...]

Issue #178, opened 2020-09-15
[ENH] Figure out QC metrics for reconstruction outputs


mattcieslak:
It would be enormously useful to have QC metrics for reconstruction outputs for sanity checking and pipeline benchmarking.

One great option are some of the fib qc metrics from DSI Studio. @frankyeh are there any of DSI Studio's fib QC measures that can be accessed via the commandline interface?

frankyeh:
There is only one for QSDR reconstructed FIB files: the R squared value
between the subject image and template image. It can be accessed by
unzipping the fib.gz file, renaming it as .mat, and loading it as a MATLAB
file.
Another is the neighboring DWI correlation for SRC files.
Frank

mattcieslak:
What about the algorithm that calculates:

```
Fiber coherence index: 946058
Fiber discoherent index: 115522
```

Would this be suitable as a qc metric?

frankyeh:
Yes, but it may not be as good as NDC. and thus I haven't put it on the
command line yet.
Frank

mattcieslak:
If you're interested, I'd be happy to benchmark this on a lot of data! I really like that it works on any reconstruction

frankyeh:
Sure. I will list it on my higher priority list and should be ready in the
command line in days.
Frank

frankyeh:
updated.
how DSI Studio supports qc on fib
dsi_studio --action=qc --source=my.fib.gz

mattcieslak:
@frankyeh you are amazing! I will get in touch off github about possible benchmarking strategies.

mattcieslak:
@frankyeh did the command line call for calculating QC change with this update?

frankyeh:
Should still be the same. If not, I will fix it.
Frank

mattcieslak:
@frankyeh the directory where the results file is written changed, no need to fix anything. I am noticing some different behavior in how bvecs are read. Is there any reason that the newest DSI Studio might be flipping the y direction even with --check-btable=0?

dsi_studio from January:
![image](https://user-images.githubusercontent.com/170026/95112730-5496b980-070f-11eb-8a4c-04c28a445462.png)

dsi_studio from last week:
![image](https://user-images.githubusercontent.com/170026/95112840-7a23c300-070f-11eb-97cd-641655d6c3cd.png)

frankyeh:
Yes, I received a request to flip y if DSI Studio reads b-table from FSL's
bvec because of the orientation convention differences.
[...]

PR #181, opened 2020-09-20
[ENH] Update FSL, DSI Studio, add SimpleITK

mattcieslak:
update some dependencies for bug fixes (#180) and new features (#178) and add a new dependency (#177)


mattcieslak:
Update Eddy inputs for new features #184

mattcieslak:
@richford which version of numpy would you recommend?

richford:
pyAFQ uses `numpy=1.17.5`, so if all other things are equal, I'd recommend that.

PR #190, opened 2020-10-27
DEP: Bump numpy version

### LINKED ISSUE #189: Numpy version in Dockerfile remains at 1.15.4
The conversation in #181 seemed to indicate the the numpy version should be bumped to 1.17.5, however the Dockerfile still shows 1.15.4. If I understand correctly, the numpy version should be bumped.

richford:

This PR bumps the numpy version in the Dockerfile to 1.17.5. I believe this was mostly accomplished in #181 but I noticed the old numpy version in the Dockerfile, so this PR is an attempt to remedy that.

Resolves #189


mattcieslak (commented):
This will also need to go in setup.cfg

richford:
@mattcieslak, I believe that last commit will fix the circleci failures (time will tell). However, I'm also getting notifications of a travis failure [here](https://travis-ci.com/github/richford/qsiprep/builds/193410188). Should I worry about the flake8 failures there? They don't seem to be from changes that I introduced.

mattcieslak:
Ah, sorry about the flake8, I switched to vs code and forgot to enable a linter. I'll fix these and push

richford:
Thanks @mattcieslak. I also don't understand the source of the circleci failures above. They seem to indicate that graphviz is not installed. But it's passing in the other circleci checks so the failure mode is intermittent. They might pass if if we simply rerun those two jobs but I don't have permissions to do that.

mattcieslak:
Found the issue:

```
PackagesNotFoundError: The following packages are not available from current channels:

  - numpy=1.17.5
```

Have you ever installed this exact version through conda?

PR #203, opened 2020-12-16
[WIP] Patch2Self Denoiser 

ShreyasFadnavis:
## Adds Patch2Self  
Hi Everyone,

This PR adds the basic version of our method: `Patch2Self: Denoising Diffusion MRI with Self-supervised Learning` 

The paper was published in NeurIPS 2020. Link: [Paper](https://papers.nips.cc/paper/2020/hash/bc047286b224b7bfa73d4cb02de1238d-Abstract.html)

- [x] Make Interface via DIPY (added to dipy.py)
- [x] Add to workflows
- [x] Add tests

@mattcieslak Does this look okay? Let me know if I missed something!
Where would this workflow ideally go?


mattcieslak:
Thanks for getting this started! Which version of dipy has the current stable implementation of Patch2Self?

ShreyasFadnavis:
> Thanks for getting this started! Which version of dipy has the current stable implementation of Patch2Self?

`1.4.0+` will have Patch2Self  👍🏽

mattcieslak:
Is that version available on pypi?

arokem:
Not yet. @skoudoro : when is DIPY 1.4.0 scheduled for release?

A potential stop-gap would be to release the prototype (
https://github.com/ShreyasFadnavis/patch2self) to pypi and then replace
that when DIPY 1.4.0 is released.

mattcieslak:
If patch2self can work as a standalone package that would make it much easier to get into qsiprep. There is a delicate and old combination of vtk and fury that works to make the odf images in the html reports. If patch2self doesn't depend on dipy it would be awesome to be able to install it by itself

skoudoro:
> Not yet. @skoudoro : when is DIPY 1.4.0 scheduled for release?

DIPY 1.4.0 Release expected December 28th if no surprise.




For your information, all DIPY workflow/command-line are available in Nipype. No need to do them again. When you import Nipype, it checked your DIPY version and the available workflows, and then convert DIPY workflow to the Nipype interface. You can look at the code [here at `dipy_to_nipype` function](https://github.com/nipy/nipype/blob/master/nipype/interfaces/dipy/base.py)  and [here for the import](https://github.com/nipy/nipype/blob/master/nipype/interfaces/dipy/reconstruction.py#L26-L33) 

So in this case. 

`from nipype.interfaces.dipy.preprocess import Patch2Self`

for a reconstruction method (syntax to check):

`from nipype.interfaces.dipy.reconstruction import ReconstDti,  ReconstMAPMRI, ReconstCSD, ReconstCSA, etc...`

for tractography

`from nipype.interfaces.dipy.tracking import PFTrackingPAM, etc...`

I know, it is not well documented, but feel free to try them. I hope it helps.

ShreyasFadnavis:
> If patch2self can work as a standalone package that would make it much easier to get into qsiprep. There is a delicate and old combination of vtk and fury that works to make the odf images in the html reports. If patch2self doesn't depend on dipy it would be awesome to be able to install it by itself

 I can make this happen too (quite simple actually)! I can make a package on my repo and then we can see how the `DIPY` version can be integrated 👍🏽

ShreyasFadnavis:
Hi @mattcieslak ! After brainstorming with @Garyfallidis we came to a workaround where `Patch2Self` can be just added to the codebase of QSIPrep until the newer version of `DIPY` is supported. The code is very modular, i.e. just 1 file, without any dependencies on other `DIPY` modules. Does that sound like a reasonable stopgap?

mattcieslak:
That sounds great! You can add it in where you're already working and I will add in the commandline options to use it. Thanks!

ShreyasFadnavis:
> That sounds great! You can add it in where you're already working and I will add in the commandline options to use it. Thanks!

Should I create a new file named `patch2self.py` or do you want me to add the code to `dipy.py` within `interfaces`?

mattcieslak:
How about putting the patch2self code in qsiprep.interfaces.patch2self and having the qsiprep.interfaces.dipy.Patch2Self class call it? Than when the algorithm gets into Dipy or its own package we can just switch the import

ShreyasFadnavis:
> How about putting the patch2self code in qsiprep.interfaces.patch2self and having the qsiprep.interfaces.dipy.Patch2Self class call it? Than when the algorithm gets into Dipy or its own package we can just switch the import

On it! Thanks @mattcieslak 👍🏼

qsiprep/workflows/dwi/pre_hmc.py
```diff
@@ -121,37 +123,37 @@ def init_dwi_pre_hmc_wf(scan_groups,
...
+                                               dwi_denoise_window=dwi_denoise_window,
+                                               unringing_method=unringing_method,
+                                               dwi_no_biascorr=dwi_no_biascorr,
+                                               denoise_method=denoise_method,
+                                               no_b0_harmonization=no_b0_harmonization,
+                                               denoise_before_combining=denoise_before_combining,
+                                               orientation=orientation,
+                                               omp_nthreads=omp_nthreads,
+                                               source_file=plus_source_file,
+                                               phase_id=pe_axis + "+ phase-encoding direction",
```
ShreyasFadnavis, line 137: Hi @mattcieslak : In your commit you hadn't added this line. I added it to resolve conflicts with `master`. Is that okay?

requirements.txt
```diff
@@ -5,3 +5,4 @@ dipy==1.0.0
 svgutils
 nipype==1.2.0
 sentry_sdk==0.13.1
+scikit-learn==0.21.0
```
ShreyasFadnavis, line 8: I just added `0.21.0` so that if people who have even older versions of `sklearn` are okay. We can switch to `0.23.0` which seems to be the recent stable release too 👍🏽

mattcieslak:
@ShreyasFadnavis What is the preferred way to calculate a 3D summary of the amplitude of the noise? `dwidenoise` has a `--noise-image` option, which I believe is just the RMS of the noise over all the volumes

qsiprep/workflows/dwi/pre_hmc.py
mattcieslak: this is ok

ShreyasFadnavis:
> @ShreyasFadnavis What is the preferred way to calculate a 3D summary of the amplitude of the noise? `dwidenoise` has a `--noise-image` option, which I believe is just the RMS of the noise over all the volumes

@mattcieslak : Yes! We also follow a similar strategy. I generally compute an RMS error (residuals) per gradient direction.  
`rms_diff = np.sqrt((noisy_img - denoised_img) ** 2)`

So we can return a `4D` noise image instead!

requirements.txt
mattcieslak: it was previously set to 0.20.2. https://github.com/PennBBL/qsiprep/blob/master/Dockerfile#L263 Is this too old?

ShreyasFadnavis: No no, I think its completely fine!

ShreyasFadnavis:
> > @ShreyasFadnavis What is the preferred way to calculate a 3D summary of the amplitude of the noise? `dwidenoise` has a `--noise-image` option, which I believe is just the RMS of the noise over all the volumes
> 
> @mattcieslak : Yes! We also follow a similar strategy. I generally compute an RMS error (residuals) per gradient direction.
> `rms_diff = np.sqrt((noisy_img - denoised_img) ** 2)`
> 
> So we can return a `4D` noise image instead!

@mattcieslak : I added this functionality from within `Patch2Self`. Let me know if this is not suitable, I can always revert changes!

ShreyasFadnavis:
@mattcieslak Do you have an example dataset I can test this on? How do you generally go about doing sanity checks?

mattcieslak:
@ShreyasFadnavis I'm testing this locally tonight and should have a working version soon

ShreyasFadnavis:
> @ShreyasFadnavis I'm testing this locally tonight and should have a working version soon

Hey, let me know if you face any issues 👍🏽 It could have happened that I broke something when porting! I can fix it ASAP :)

mattcieslak:
Looks like it's working! Here is the report from using [patch2self](https://5641-156589095-gh.circle-artifacts.com/0/tmp/DSDTI/derivatives/qsiprep/sub-PNC.html) and here is the same data using [mp-pca](https://5646-156589095-gh.circle-artifacts.com/0/tmp/DSDTI/nofmap_derivatives/qsiprep/sub-PNC.html).

I still need to fix the boilerplate to cite you correctly, but was also wondering what you would recommend as defaults for the parameters. For dwidenoise I have the default window size set to (5, 5, 5).

ShreyasFadnavis:
> Looks like it's working! Here is the report from using [patch2self](https://5641-156589095-gh.circle-artifacts.com/0/tmp/DSDTI/derivatives/qsiprep/sub-PNC.html) and here is the same data using [mp-pca](https://5646-156589095-gh.circle-artifacts.com/0/tmp/DSDTI/nofmap_derivatives/qsiprep/sub-PNC.html).
> 
> I still need to fix the boilerplate to cite you correctly, but was also wondering what you would recommend as defaults for the parameters. For dwidenoise I have the default window size set to (5, 5, 5).

Hi @mattcieslak ! Thank you for this 👍🏽 For `patch2self` the default `patch_radius` should be `[0, 0, 0]`. If the number of gradient directions is less than 20, then increasing the `patch_radius` to `[1, 1, 1]` = `patch_size` of `[3, 3, 3]` is required (very rare for connectome data -- required for `spinal cord` or other body parts). This is the best part! We discovered that if the number of gradient directions is more than `50`, which is generally the case, spatial approximation does not add any value.

Users don't need to tune any parameter! 💯 

Maybe we can document somewhere that: if the number of gradient directions is less, increase the patch radius?
(perhaps raise a warning)

ShreyasFadnavis:
@mattcieslak 1 more thing: `patch2self` can be applied at any point in the pipeline... So can we try it after motion correction?

mattcieslak:
That should probably get its own PR, since it would require some larger changes to the overall workflow.

ShreyasFadnavis:
> That should probably get its own PR, since it would require some larger changes to the overall workflow.

Sounds good! Will follow up with a separate PR for that!

ShreyasFadnavis:
Hi @mattcieslak, is this good to go? Do you need something else on my end?

PR #214, opened 2021-01-13
[ENH] improve dsi_studio qc interfaces and remove hard link dependency

octomike:

Two days ago we tripped over the fact that the use of [`os.link()` in dsi_studio.py](https://github.com/PennBBL/qsiprep/blob/master/qsiprep/interfaces/dsi_studio.py#L135) fails for file systems without or with limited hard link support ([BeeGFS](https://www.beegfs.io/wiki/FAQ#hardlinks) in our case). A quick fix would have been to just wrap it into a try-except block and use `shutil.copyfile()` on failure. However, I was curious what was going on in the interfaces and I found that the reason a tmp directory was created (at least for [DSIStudioSrcQC](https://github.com/PennBBL/qsiprep/blob/master/qsiprep/interfaces/dsi_studio.py#L106) was probably caused by some decision in the [DSI-Studio src](https://github.com/frankyeh/DSI-Studio/blob/ea38389afb12877211ad5bf9f2eb1630d0b33cc1/main.cpp#L74) to not support symbolic links when searching for input source files in a directory.

It turns out that we probably don't need to create the tmp directory if we only run the qc action on single files with `dsi_studio --action=qc --src...`. The symbolic link check does not fire then. This allowed for some minor code deduplication and simplification. Specifically, I merged `DSIStudioSrcQC` and `DSIStudioFibQC` into a single `DSIStudioQC` interface as they both just run the same dsi_studio command on different (single) input files.

Note that this is my first time playing around with traited specs and nipype interfaces, so I probably did something wrong :smile:  Still, I think this is a non-invasive improvement and at least a bug fix for file systems without hard link support.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -29,6 +29,11 @@
       "affiliation": "Perelman School of Medicine, University of Pennsylvania, PA, USA",
       "name": "Satterthwaite, Theodore D.",
       "orcid": "0000-0001-7072-9399"
+    },
+    {
+      "name": "Krause, Michael",
+      "affiliation": "Max Planck Institute for Human Development, Berlin, Germany",
+      "orcid": "0000-0002-3878-6542"
     }
   ],
   "keywords": [
--- a/qsiprep/interfaces/dsi_studio.py
+++ b/qsiprep/interfaces/dsi_studio.py
@@ -87,61 +87,48 @@ class DSIStudioCreateSrc(CommandLine):
         return outputs
 
 
+class _DSIStudioQCOutputSpec(TraitedSpec):
+    qc_txt = File(exists=True, desc="Text file with QC measures")
+
+
+class DSIStudioQC(SimpleInterface):
+    output_spec = _DSIStudioQCOutputSpec
+
+    def _run_interface(self, runtime):
+        # DSI Studio (0.12.2) action=qc has two modes, depending on wether the
+        # input is a file (src.gz|nii.gz)|(fib.gz) or a directory. For
+        # directories, the action will be run on a number of detected files
+        # (which *cannot* be symbolic links for some reason).
+        src_file = fname_presuffix(self.inputs.src_file, newpath=runtime.cwd)
+        cmd = ['dsi_studio', '--action=qc', '--source='+src_file]
+        proc = Popen(cmd, cwd=runtime.cwd, stdout=PIPE, stderr=PIPE)
+        out, err = proc.communicate()
+        if out:
+            LOGGER.info(out.decode())
+        if err:
+            LOGGER.critical(err.decode())
+        self._results['qc_txt'] = src_file.replace(self.ext, '.qc.txt')
+        return runtime
+
+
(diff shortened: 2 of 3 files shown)


mattcieslak:
Thanks for your contribution! I will read this over and send some comments soon. 

The DSI Studio interfaces have a strange setup because they write their outputs into directories relative to their input (there is no option for what the output file will be named). So to get the output file in the right directory we would need to symlink the inputs into specific locations.

octomike:
> So to get the output file in the right directory we would need to symlink the inputs into specific locations.

Yes, I am still a little confused about how nipype would create such symlinks. I haven't confirmed this yet, but is it possible that the `copyfile` argument to the File trait has 3 states? It seemed to me that if this is `not set` at all, the input will point to the exact location of the connected file. If it is set to `False` it appears there is a new symlink created inside the current directory (that's what we need) and (also not checked) `True` would probably create a full copy in the current dir. This is why I [set](https://github.com/PennBBL/qsiprep/pull/214/files#diff-d7cf0c3ff84f8dad28d621a519759b5fdf66b99906a0b82c84d4f09dd1e9096bR125) `copyfile=False` here.

octomike:
Carefully checking in here. Is there anything I could do to improve this PR?

mattcieslak:
Sorry for the delay! There is some behavior in DSI Studio where it only accepts an input file and doesn't let you specify where the output goes. So to get around this I've been symlinking the input file into the sandbox directory for the interface's run and passing that to DSI Studio so the resulting text file gets added to the right place.

Are you using qsiprep from a container? Or are you using a local installation?

octomike:
> Sorry for the delay! 

Not to worry, please!

> There is some behavior in DSI Studio where it only accepts an input file and doesn't let you specify where the output goes. So to get around this I've been symlinking the input file into the sandbox directory for the interface's run and passing that to DSI Studio so the resulting text file gets added to the right place.

I found that if the input is a single file and not a directory, DSI studio can work with symlinked input files. Before the interface created hard(!) links in a temp directory. The hard link breaks on our distributed HPC file system (the initial issue for us) and the temp directory is not really necessary I think. That's why I decided to simplify the interface a little and it also allowed for some deduplication.

> Are you using qsiprep from a container? Or are you using a local installation?

Yes, we are running qsiprep with singularity and aim to follow the main releases from docker-hub. Right now we use a patched container, because as it is now, the DSI studio interface is incompatible with our system.

mattcieslak:
This came up again in #244. Could you please merge and rebase from the current master branch?

octomike:
Sure thing - I rebased on master and force pushed to the feature branch. Hope that's okay?

mattcieslak:
perfect- thanks!

mattcieslak:
This is great, thanks again for the PR.  Could you please add your info to the .zenodo.json file so you can be added to the author list?

octomike:
Feels weird for a minimal change like this, but if you insist :blush:  (force pushed)

PR #223, opened 2021-02-06
[ENH] Make recon workflows correctly process dwi-only inputs

mattcieslak:
Currently it is assumed that there is a T1w brain mask available going in to the reconstruction workflows. This breaks if the user picked --dwi-only. 

TODO:

- [x] DSI Studio
- [ ] Dipy
- [ ] Mrtrix
- [ ] Mrtrix ss3t


mattcieslak:
Closed after refactor in 0.15

PR #229, opened 2021-02-19
FIX: remove cmake installer after compiling ANTs

cookpa:

The file /cmake-3.11.4-Linux-x86_64.sh is used for compiling ANTs but is left in the container root directory. Removing it saves 33 Mb.

None


mattcieslak:
Nice catch! Thank, Phil

PR #239, opened 2021-03-19
[ENH] Change patch2self defaults

mattcieslak:
Changes patch2self defaults based on https://github.com/dipy/dipy/issues/2334


ShreyasFadnavis:
Thanks a lot @mattcieslak for getting at this super fast! Also thank you @arokem for pointing this out!

PR #257, opened 2021-06-15
[ENH] Add NODDI/AMICO recon workflows

mattcieslak:
Add the AMICO package so qsiprep can reconstruct the NODDI model

TODO:
-------

 - [x] Add reorient-to-fsl workflow so vectors show up correctly in fsleyes
 - [x] Add an interface for NODDI

Refer to https://github.com/cookpa/amico-noddi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -356,4 +356,5 @@ RUN  mkdir -p /sngl/data \
   && mkdir /sngl/scratch \
   && mkdir /sngl/spec \
   && mkdir /sngl/eddy \
+  && mkdir /sngl/filter \
   && chmod a+rwx /sngl/*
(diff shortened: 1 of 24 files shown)


Dockerfile
```diff
@@ -318,6 +318,9 @@ RUN echo "${VERSION}" > /src/qsiprep/qsiprep/VERSION && \
 RUN python -c "from matplotlib import font_manager" && \
     sed -i 's/\(backend *: \).*$/\1Agg/g' $( python -c "import matplotlib; print(matplotlib.matplotlib_fname())" )
 
+RUN mkdir -p ${HOME}/.dipy
+RUN python -c "import amico; amico.core.setup()"
```
cookpa, line 322: This is going to write to /home/qsiprep/.dipy, but Singularity users will have their own $HOME. I got around this by exporting DIPY_HOME in the Dockerfile

mattcieslak: Earlier in the Dockerfile it creates a unique $HOME, this turned out to be ok

PR #268, opened 2021-07-07
[ENH] Ingress HCP preprocessed data

mattcieslak:
HCP preprocessed data has comparable quality to QSIPrep-preprocessed data. Since so much HCP preprocessed data is available, we should be able to load it directly into the reconstruction workflows.


liuzhenqi77:
Hi, just wondering are there any updates on this? Thanks!

Issue #277, opened 2021-07-22
Fails to find DWI pairs


aghayoor:
Hi developer team,

I am trying to run QSIPrep on some PPMI data.
In their acquisition set up, they have acquired two sessions of DWI data in opposite phase encoding direction, that can be used for susceptibility artifact correction. For this, I used `--distortion-group-merge average` argument in qsiprep.

Here is the full command line I used to run qsiprep:

> qsiprep-docker $datadir/BIDS/ $outdir participant --participant-label 01 --output-resolution 1.2 --fs-license-file ./fs_license.txt --distortion-group-merge average --verbose

However, I receive this error:

>  Node: qsiprep_wf.single_subject_01_wf.sub_01_final_merge_wf.distortion_merger
>  Working directory: /tmp/work/qsiprep_wf/single_subject_01_wf/sub_01_final_merge_wf/distortion_merger
> 
>  
>  Node inputs:
>  
>  b0_refs = <undefined>
>  b0_threshold = 100
>  bids_dwi_files = <undefined>
>  bval_files = <undefined>
>  bvec_files = <undefined>
>  carpetplot_data = <undefined>
>  denoising_confounds = <undefined>
>  dwi_files = <undefined>
>  harmonize_b0_intensities = True
>  original_bvec_files = <undefined>
>  raw_concatenated_files = <undefined>
>  verbose = False
>  
>  Traceback (most recent call last):
>    File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
>      result["result"] = node.run(updatehash=updatehash)
>    File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
>      result = self._run_interface(execute=True)
>    File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
>      return self._run_command(execute)
>    File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
>      result = self._interface.run(cwd=outdir)
>    File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 428, in run
>      runtime = self._run_interface(runtime)
>    File "/usr/local/miniconda/lib/python3.7/site-packages/qsiprep/interfaces/dwi_merge.py", line 168, in _run_interface
[...]

PR #279, opened 2021-07-28
[FIX] handle case where there is only 1 b=0

mattcieslak:
the concatenation strategy for merging distortion groups ran into a bug when only 1 b=0 image was present in a distortion group (#277). This should fix concat for this situation


cookpa:
Does 1 b=0 also apply to sbrefs? The data I saw the error with had multiple b=0 in each DWI series, but a corresponding sbref for each.

mattcieslak:
The sbrefs have always been ignored, maybe this is coming from one of the epi fieldmaps? The code that caused the error on concat definitely only had a 3d volume in it. I might need to debug this on some actual data

giulia-berto:
I'm also affected by this, i.e. I have only 1 b0 image (3D) and the pipeline is failing when merging distortion groups. Is it possible to accept this PR?

PR #293, opened 2021-09-16
Crucial Fixes for Patch2Self

ShreyasFadnavis:
Fixes parameters upstream with `DIPY's Patch2Self`.

This is the reason why @SendyCaffarra was getting a bug. This issue was fixed in `Patch2Self` for DIPY in #2334 

[[PR Link]( https://github.com/dipy/dipy/issues/2334)]

Also setting default `'model='ols'`

I am also taking this opportunity to fix the naming [MP-PCA --> Generic "DWI Denoising"] in the reports.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -255,8 +255,8 @@ jobs:
       - run:
           name: Install graphviz
           command: |
-              sudo apt-get update
-              sudo apt-get install -y graphviz libblas-dev \
+              sudo apt update
+              sudo apt install -y graphviz libblas-dev \
                liblapack-dev gfortran
       - run:
           name: Install qsiprep.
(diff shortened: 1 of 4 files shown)


ShreyasFadnavis:
@mattcieslak can you take a look at this? 

At some point, we should just depend on DIPY 1.4.1 or greater and use Patch2Self from there since now it is released and on master. This would avoid issues like these in the future :)

ShreyasFadnavis:
Looks all green 💯 ! Let me know if anything else is needed on this PR 👍🏽

ShreyasFadnavis:
1 quick question -- unrelated to this PR. In the QSIPrep report -- why do we show contours on the denoised output? Is there a specific reason?

I am asking because it just makes the denoising difference much less visible.

PR #303, opened 2021-09-30
[ENH] Add pyafq

mattcieslak:
This PR adds PyAFQ as a Reconstruction workflow.


36000:
I added the ability to generate a qsiprep-compatible json pipeline descriptor to the pyAFQ API, so if new versions add new parameters, we can just use this tool. I am attaching a rough draft here.
[pyafq_full.txt](https://github.com/PennLINC/qsiprep/files/7262959/pyafq_full.txt)

mattcieslak:
Looks like trying to add pyAFQ to the dependencies in setup.cfg caused the build to crash. In your dockerfile do you install pyAFQ before or after qsiprep?

36000:
I install qsiprep then pyAFQ. However I use "--force-reinstall" for pyAFQ and the only dependency conflict came from matplotlib (we use seaborn which wants matplotlib >=3, but its not crucial to the main part of the pipeline, I should move that to optional dependencies). It looks like here it is taking too long for pip to determine what versions of each of our dependencies to install. I am not sure what the best practice is here for resolving that in a way where we don't have to worry about it when we update the dependencies of each of our packages.

36000:
@mattcieslak  we have simplified the dependencies for pyAFQ, re-running this may solve the problem

mattcieslak:
Is there a specific version I should specify in setup.cfg or just the latest?

36000:
pyafq==0.9
This is the latest version

mattcieslak:
@36000 in the line

```
parallel_segmentation": "{'n_jobs': -1, 'engine': 'joblib', 'backend': 'loky'}",
```

will this cause pyafq to use all the available cpus on the host?

mattcieslak:
excellent! We can use the number of cpus specified by the user on the commandline here. It also looks like the dependencies are compatible now, which is awesome.

Question - what does pyAFQ do if there is more than image in the qsiprep results directory? Currently the reconstruction workflows get applied to every preprocessed dwi file found in the qsiprep outputs

36000:
This is something we can discuss in the meeting tomorrow, but there are some parameters I did not put in the config.json, which we will set directly:

`# The path to preprocessed diffusion data organized in a BIDS
# dataset. This should contain a BIDS derivative dataset with
# preprocessed dwi/bvals/bvecs.
bids_path = ""

# Filter to pass to bids_layout.get when finding DWI files.
# Default: {"suffix": "dwi"}
bids_filters = "{'suffix': 'dwi'}"

# The name of the pipeline used to preprocess the DWI data.
# Default: "all".
preproc_pipeline = "all"

# list of participant labels (subject IDs) to perform
# processing on. If None, all subjects are used.
# Default: None
participant_labels = ""

# path to output directory. If None, outputs are put
# in a AFQ pipeline folder in the derivatives folder of
# the BIDS directory. pyAFQ will use existing derivatives
# from the output directory if they exist, instead of recalculating
# them (this means you need to clear the output folder if you want
# to recalculate a derivative).
# Default: None
output_dir = ""

# BIDS filters for inputing a user made tractography file.
# If None, tractography will be performed automatically.
# Default: None
custom_tractography_bids_filters = ""

# This will be used to create
# the brain mask, which gets applied before registration to a
# template.
# If None, no brain mask will not be applied,
# and no brain mask will be applied to the template.
# Default: B0Mask()
brain_mask = "B0Mask({})"` 

To answer your question, we should use `bids_filters` to specify which file to use, otherwise pyAFQ will just choose one that most closely matches the filters and throw a warning if more than one match.

36000:
@mattcieslak Hello! Sorry for the delay on this PR, I got distracted by other things. I have made some changes and rebased, but I have two problems:
1. I cannot push to this PR, and I would need to force push as well due to the rebase
2. How can I run tests locally? I made my own PyAFQReconTest based off of the AMICOReconTests.sh, but when I try to run it I get the error:
```
Error response from daemon: error while creating mount source path '/Users/mcieslak/projects/qsiprep/qsiprep': mkdir /Users/mcieslak: operation not permitted.
```
Overall I think I am close to being ready to merge, but I want to be able to run tests locally to make sure I understand everything that's going on.

36000:
moved to #398 , this can close now

Issue #313, opened 2021-10-25
Transformation from original T1w to qsiprep space


Giuseppe1988:
Dear experts,
I know that this issue has been partially addressed in previous topics but I can't seem to find the definitive solution for this.

I know that the spatial transformation mapping the original T1w to the qsiprep space (the one in which the desc-preproc_T1w and space-T1w_desc-preproc_dwi are) is not saved in the derivatives, but could you please specify the exact process (and the corresponding transformations in the work directory) to replicate the original T1w to qsiprep_T1w step.

Thank you so much in advance for your help.

Best,
Giuseppe

Issue #305, opened 2021-10-05
[feature request] Save out T1/b0-to-ACPC transform to derivatives


smeisler:
Transform depends on if anatomical workflow is being used (T1) or not (b0).

It can be found in`/work/qsiprep_wf/single_subject_XX_wf/dwi_preproc_wf/b0_anat_coreg/itk_to_rigid/`.

mattcieslak:
This is in the derivatives now

PR #314, opened 2021-10-25
[ENH] add datasink for acpc transform

mattcieslak:
Adds a datasink for the T1w-to-acpc transform file (#313, #305)


mattcieslak:
Yes, thanks!

PR #332, opened 2022-01-14
DOC: How to avoid some pitfalls of using filters

cookpa:

Updating the preprocessing documentation to note two things:

1. If you are using BIDS filters to process a subset of the data, be careful that your subsequent calls don't overwrite subject-level output. For example, if you have two time points and you want to process them separately, you might call with a filter to get session MR1 and again with a filter to get session MR2. But the T1w processing for MR2 will overwrite that for MR1, unless you send the output to different directories.

2. Clarify that not all entities known to PyBids can be used in filters. The link in the doc to the master list includes entities for all BIDS data, MRI or not. The filters will not work if you use an invalid entity (we found this in the FTDC by inadvertently using an entity "proc-", for MEG data, instead of "rec-", for MRI).


cookpa:
Sorry @mattcieslak, I don't know why all these other commits are in the PR. I just wanted e08faf3! Will try to fix later

Issue #325, opened 2021-12-10
fs-license error when running qsiprep


andrew-yian-sun:
I am new to qsiprep. I ran the following code:
```
sudo docker run -ti --rm 
--volume /blc/NIA_R21/data_mri/dataOLD_BIDS_test_0061:/data:ro 
--volume /blc/NIA_R21/data_mri/dataOLD_qsiprep_test:/out 
pennbbl/qsiprep:latest 
/data 
/out 
participant 
--output_resolution 0.95 
--skip_bids_validation 
--fs-license-file $FREESURFER_HOME/license.txt
```
However, I get the following error message:
```
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/qsiprep", line 8, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.7/site-packages/qsiprep/cli/run.py", line 578, in main
    set_freesurfer_license(opts)
  File "/usr/local/miniconda/lib/python3.7/site-packages/qsiprep/cli/run.py", line 539, in set_freesurfer_license
    (for free) by registering at https://surfer.nmr.mgh.harvard.edu/registration.html""")
RuntimeError: ERROR: a valid license file is required for FreeSurfer to run. fMRIPrep looked for an existing license file at several paths, in this order: 1) command line argument ``--fs-license-file``; 2) ``$FS_LICENSE`` environment variable; and 3) the ``$FREESURFER_HOME/license.txt`` path. Get it (for free) by registering at https://surfer.nmr.mgh.harvard.edu/registration.html
Sentry is attempting to send 1 pending error messages
Waiting up to 2 seconds
Press Ctrl-C to quit
```
The license.txt file exists in `$FREESURFER_HOME`. I even replaced it with a new one thinking maybe the old one was for the wrong OS, but still got the same error message. I get the same error message whether I specify the license file with the `--fs-license-file` or not.

Details:
```
sudo docker version
Client: Docker Engine - Community
 Version:           19.03.5
```
OS: CentOS 7

Can anyone help me with this issue?

cookpa:
One thing to check is if your OS is in FIPS mode. FreeSurfer will not recognize its license files on a system with FIPS, because its encryption algorithm doesn't work in such environments.

andrew-yian-sun:
Thanks for the fast reply. 
I used `cat /proc/sys/crypto/fips_enabled` and the output was `0`, so I don't think FIPS is enabled.

pcamach2:
[...]

PR #355, opened 2022-02-24
ENH: Explain mounting Freesurfer license.txt to Docker & Singularity

pcamach2:
Examples of mounting the Freesurfer license.txt given to prevent issues similar to #325

Changed the installation page of the docs to include examples of mounting the Freesurfer license.txt given to prevent issues running Freesurfer workflows in Docker or Singularity containers of QSIPrep - similar to #325 (closed with the method used in the added examples).

Since no change was made to the qsiprep tool itself in this PR, changes to documentation consist of the added example run commands. I think anyone choosing the option of running Freesurfer here likely already knows how to get the necessary license for their work, but it might be worth adding a link if not.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -41,6 +41,11 @@
       "orcid": "0000-0001-7072-9399"
     },
     {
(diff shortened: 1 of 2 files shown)


mattcieslak:
Thank you for these additions!!

Issue #206, opened 2020-12-22
Add DKI, DKI-Micro, MSDKI and SMT2 from DIPY


ShreyasFadnavis:
It would be great to add Diffusion Kurtosis Imaging (DKI) and its variants from `DIPY`! 
A bunch of different measures can be derived from these models 💯 

@arokem what do you think? I can take the lead on this if that's okay 👍🏽

arokem:
I think that it would be fantastic to have these available to qsiprep users! 

It does add two levels of complexity: the DKI model only works in multi-shell acquisitions (that's probably already sorted out here for other things), and it adds more configuration knobs that need to be set.

mattcieslak:
I like this idea. @arokem or @ShreyasFadnavis do you know of a python-based sampling scheme classifier? 

I was thinking about adding a "ModelMix" reconstruction workflow that would attempt to detect what sampling scheme was used and then calculate an assortment of  scalar values from each.

arokem:
We a couple of functions you could use to classify multi- vs. single-shell [here](https://github.com/dipy/dipy/blob/master/dipy/core/gradients.py#L793) and [here](https://github.com/dipy/dipy/blob/master/dipy/core/gradients.py#L714)

mattcieslak:
these were added!

PR #366, opened 2022-03-03
[ENH] Add DKI

mattcieslak:
Adds a workflow for DKI (partial #206)


qsiprep/interfaces/dipy.py
```diff
@@ -645,6 +645,37 @@ class TensorReconstructionOutputSpec(DipyReconOutputSpec):
     cnr_image = File()
 
 
+class TensorReconstruction(DipyReconInterface):
```
arokem, line 648: Should this really be `TensorReconstruction`? I might be misunderstanding something, but it seems to be over-ridden on line 679?

qsiprep/interfaces/dipy.py
```diff
@@ -675,3 +706,73 @@ def _run_interface(self, runtime):
             self._results[metric + "_image"] = out_name
 
         return runtime
+
+
+class _KurtisisReconstructionInputSpec(DipyReconInputSpec):
```
arokem, line 711: Is this spelling of "Kurtosis" intentional?

arokem (commented):
I am excited to have this here! So I took a look. I noticed a couple of things in the code, so even though it's merged, I have a couple of questions.

PR #405, opened 2022-04-23
BUG: Prevent divide by zero if bvec is all zeros

### LINKED ISSUE #375: Question about Preprocessing HCP-style
I have a dataset that has full dwi scans in both AP and PA directions. My understanding is that I should be including the `--distortion-group-merge average` option in command line, per the "Preprocessing HCP-style" section. The documentation here suggests I should be including the `--combine-all-dwis` option as well, but I believe that flag has been removed (and is the default option?).

However, in terms of setting up the BIDS-valid dataset, am I right in thinking that I should put both of these scans (AP and PA) in the dwi folder? Also, is there any metadata field that needs to be added? I have tried a couple different options but kept running into errors.

Thanks,
Andrew

cookpa:
Some older data has bvals / bvecs of exactly 0, because the don't account for the non-diffusion gradients that make the real diffusion weighting slightly non-zero.

average_bvec now checks if gradients are identical first, if they are it just returns the answer without doing any of the normalization that leads to division by zero.

Fixes #375.

### DIFF
--- a/qsiprep/interfaces/dwi_merge.py
+++ b/qsiprep/interfaces/dwi_merge.py
@@ -365,6 +365,11 @@ def average_carpetplots(carpet_list, image_pairs):
 
 
 def average_bvec(bvec1, bvec2):
+    # return straight away if the bvecs are identical
+    # This prevents comparison of zero vectors
+    if (bvec1 == bvec2).all():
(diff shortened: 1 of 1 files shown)


mattcieslak:
Thanks again!!

PR #435, opened 2022-06-18
[WIP] ENH: Add SIFT-classic

smeisler:
Related to PennLINC/qsirecon#33.

1) `use_sift` now takes as input "1" or "2" (or `false`) to differentiate between sift versions
2) A separate SIFT1 class is created which takes in FOD, Tck and optional final streamline count (default 1million), and outputs filtered tractogram which becomes the main output of tractography fed into other workflows.

I have not updated documentation yet, but recon page would be affected


smeisler:
Hmmm, the command is being generated and executed properly, but only the un-sifted tractogram is ultimately output, and under the sift name.

mattcieslak:
This looks awesome - I'm not going to be able to code for the next few weeks, so I'll add write access for you on this repo

smeisler:
Hi @mattcieslak,

Finally returned to this after a year or so and got something working.

Here is a summary of changes:
1) `use_sift` now takes as input "1" or "2" (or false) to differentiate between sift versions
2) iFod2 now has a `sift_tracks` output_spec for SIFT1 outputs.
3) SIFT1 final streamline count can be set by specifying `term_number` in the `sift_params` in a given recon_spec. Otherwise, it will default to the default `tcksift` behavior (until streamline densities match the FOD lobe integrals).

Still in progress, might need some help with:
1) I have tested this out locally, but I imagine I should set up a test on the repo.
2) I don't think the `use_sift` options of 1, 2, or False is the most elegant, as I imagine a user could want to use SIFT2 for atlas connectivity and SIFT1 for tractography in a single recon specification.
3) Add documentation
4) Add an example recon_spec with SIFT1, perhaps something like Mrtrix-->PyAFQ spec, but starting with making a dense 10M streamline tractogram, SIFTing it to 1M, and then start PyAFQ.

Thoughts?

smeisler:
> 2\. iFod2 now has a `sift_tracks` output_spec for SIFT1 outputs.

I see now that this would be an issue for using the SIFT outputs for other workflows (e.g., PyAFQ), since they tend to look for the `tck_file` field. Will think about best way to change that.

smeisler:
@mattcieslak do you think it could make sense to have SIFT1 become a separate entity from the tractography workflow? That way, in theory one could run whole brain tractography, use SIFT2 weights for connectivity, but also be able to use a SIFT1'd tractogram for tract segmentation, all in one recon spec.

tsalo:
@smeisler given The Great Split, should this be reproduced in QSIRecon?

smeisler:
I would have to probably refactor this a lot of this given how recon specs have changed. I'm not even sure this feature would be worth adding anymore. But yes, this should be moved to qsirecon.

tsalo:
Okay I'll close this PR. Up to you if you want to open a corresponding one on QSIRecon.

PR #519, opened 2023-02-15
add DTI to shoreline

nseider:
add single tensor model to shoreline head motion correction

### DIFF
--- a/qsiprep/cli/run.py
+++ b/qsiprep/cli/run.py
@@ -376,10 +376,10 @@ def get_parser():
         '--hmc_model', '--hmc-model',
         action='store',
         default='eddy',
-        choices=['none', '3dSHORE', 'eddy', 'hmcDTI'],
+        choices=['none', '3dSHORE', 'eddy', 'tensor'],
         help='model used to generate target images for hmc. If "none" the '
         'non-b0 images will be warped using the same transform as their '
-        'nearest b0 image. If "3dSHORE", SHORELine will be used. if "hmcDTI", '
+        'nearest b0 image. If "3dSHORE", SHORELine will be used. if "tensor", '
         'SHORELine iterations with a tensor model will be used')
     g_moco.add_argument(
         '--eddy-config', '--eddy_config',
--- a/qsiprep/cli/run.py
+++ b/qsiprep/cli/run.py
@@ -376,10 +376,11 @@ def get_parser():
         '--hmc_model', '--hmc-model',
         action='store',
         default='eddy',
-        choices=['none', '3dSHORE', 'eddy'],
+        choices=['none', '3dSHORE', 'eddy', 'hmcDTI'],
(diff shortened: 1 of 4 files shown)


nseider:
@mattcieslak Here are the code changes we discussed earlier today.  Please let me know if I need to fix or clarify anything!

qsiprep/interfaces/shoreline.py
```diff
@@ -200,25 +202,38 @@ def _run_interface(self, runtime):
...
+            dti_wls = dti.TensorModel(training_gtab)
+            fit_wls = dti_wls.fit(training_data, mask=mask_array)
+            evals1 = fit_wls.evals
+            evecs1 = fit_wls.evecs
+            output_data = np.zeros(mask_array.shape)
+            for ii in range(mask_array.shape[0]):
+                for jj in range(mask_array.shape[1]):
+                    for kk in range(mask_array.shape[2]):
+                        if mask_array[ii,jj,kk] == 1:
+                            output_data[ii,jj,kk] = single_tensor(prediction_gtab, S0 = training_data[ii,jj,kk,0], evals=evals1[ii,jj,kk,:],evecs=evecs1[ii,jj,kk,:])[0]
```
arokem, line 232: This is a rather slow way of doing this (the triple loop). Might want to use the `dti_wls.predict` method, or this function, which will do this in one fell swoop: https://github.com/dipy/dipy/blob/master/dipy/reconst/dti.py#L639

qsiprep/cli/run.py
```diff
@@ -376,10 +376,11 @@ def get_parser():
         '--hmc_model', '--hmc-model',
         action='store',
         default='eddy',
-        choices=['none', '3dSHORE', 'eddy'],
+        choices=['none', '3dSHORE', 'eddy', 'hmcDTI'],
```
mattcieslak, line 379: what do you think of using just 'tensor' as the value instead of 'hmcDTI'?

nseider: I'm fine with that.  I'll change it everywhere before my next push.

nseider:
@mattcieslak Any insight to what failed in the CircleCI check?

mattcieslak:
This is a new one to me. These tests ran on the same day and didn't have  [the same error](https://app.circleci.com/pipelines/github/PennLINC/qsiprep/1284/workflows/41382a61-7694-4e23-be2f-a41e3d76d748). I just re-launched the jobs, maybe it will just go away

mattcieslak:
@nseider If you merge from the master branch the problem is fixed

nseider:
Success! Thanks for the quick fix.

mattcieslak:
This looks good to me, if it's ok with you I'll merge and add a CI test for it in a separate PR

nseider:
Works for me.  Thanks for all your help!

PR #526, opened 2023-02-22
Update Dockerfile to reflect new qsiprep-build

jbh1091:
Update version of qsiprep-build container referenced in Dockerfile from 22.10.5 to 23.2.4.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM pennbbl/qsiprep_build:22.10.5
+FROM pennbbl/qsiprep_build:23.2.4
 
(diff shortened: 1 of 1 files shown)


mattcieslak:
HA! This is the same thing I was working on in #522

jbh1091:
Oops! That's funny. I can remove this as a duplicate then, especially since I'm seeing some local errors with the new ANTs version running T1 to MNI registration (RobustMNINormalization) so this is not ready to accept. Didn't realize I needed to updated the circleci config file for testing to work properly until I saw your issue.

PR #422, opened 2022-05-27
WIP: [ENH] Pyafq visualization of the recognized bundles for reportlets

36000:
Closes #421

PR #434, opened 2022-06-18
[WIP] ENH: Export_All in pyafq

smeisler:
## Changes proposed in this pull request
I (and other pyAFQ users) might want to get the segmented bundles back, so changing the export to` export_all`.

PR #534, opened 2023-03-13
WIP: ENH: Make pyAFQ tests faster, add export all

36000:
Closes #422 and #434 by adding in the export all and choice of export functionality for pyAFQ.

I also want to reduce the runtime of the tests here. @mattcieslak , is there a way to change the parameters for one of the nodes in the json file for the test only? For example, I would like select here to be 1e4 instead of 1e6, but only when doing the circle CI tests: https://github.com/PennLINC/qsiprep/blob/80803117d8d2397b14b28668314f7e3258273447/qsiprep/data/pipelines/mrtrix_multishell_msmt_pyafq_tractometry.json#L35

### DIFF
--- a/.circleci/PyAFQReconExternalTrk.sh
+++ b/.circleci/PyAFQReconExternalTrk.sh
@@ -8,7 +8,7 @@ Reconstruction workflow tests
 All supported reconstruction workflows get tested
 
 This tests the following features:
- - Full pyAFQ pipeline
+ - pyAFQ pipeline with tractography done in mrtrix
 
 Inputs:
 -------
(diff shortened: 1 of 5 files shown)


mattcieslak:
I'd recommend adding a few lines in [here](https://github.com/PennLINC/qsiprep/blob/cd6a9605691d0c66e746b0c36d1e9020ee0bf66d/qsiprep/utils/sloppy_recon.py) so that when the `--sloppy` flag is used the parameters get changed to something lower.

mattcieslak:
@36000 I split this into two tests, one where pyafq does everything and another where the tractography is done in mrtrix beforehand. It seems like something in AFQ itself is making the test take a long time. Can you think of any shortcuts we might be able to change in the config when the --sloppy flag gets used?

arokem:
Could we potentially configure it to only find some bundles? That would shorten it substantially, I would think. 

Another option would be to skip the visualizations, which can also take some time.

mattcieslak:
I like the "only a few bundles" idea. That way we can test a few of them all the way through, including visualization. Is there a spot in the json where you can specify which bundles to do?

36000:
I changed it to do only a few bundles, hopefully also have a much faster registration, and generate only 1 visualization, so that the visualization library is tested but we don't spend too much time in it. Let's see if this test passes and if needs to be sped up more in other places.

mattcieslak:
Down to ~30 minutes! Thank you @36000! Does this test everything you'd like to test still? If so I'm going to merge

Issue #90, opened 2020-01-23
[ENH] Consider to integrate new fieldmapless distortion correction method


xiaosonghe:
There's a new development in fieldmapless distortion correction that utilize deep learning to build a b0 map from T1 data:
https://www.biorxiv.org/content/10.1101/2020.01.19.911784v1
Here's the Github page for the package:
https://github.com/MASILab/Synb0-DISCO
It seems promising and hopefully it can perform better than the current "experimental" version.

smeisler:
Of note, it is being integrated in DIPY 0.17.0 release, should make integration easier.

tsalo:
Handled by #1111.

PR #546, opened 2023-04-11
[WIP / ENH] Add Synb0-DisCo as a SYN Unwarp Method

smeisler:
Addresses #90. Add Synb0-DisCo ([paper](https://pubmed.ncbi.nlm.nih.gov/31075422/), [github](https://github.com/MASILab/Synb0-DISCO)) as an unwarping method.

SynB0-DisCo is being incorporated in the next release of DIPY, so the Python-ized version of this might be easier to incorporate into QSIPrep and provide advantages over the current ANTs implementation.

At a high level, I imagine the workflow would be something like
1) Grab the best b0 (i.e., `get_best_b0_topup_inputs_from`)
(Maybe we have to register b0 to T1?)
2) Use the b0 and T1 to synthesize a RPE DWI contrast image using Synb0-DisCo
3) Ingress that into whatever unwarping software is requested (`TOPUP, 3dQWarp, DRBUDDI`)

It would operate different from the ANTs method, which is pretty much its own workflow. This would be an intermediary step  preceding the normal unwarping methods.

I will also ping @schillkg who developed the algorithm for additional insight.


mattcieslak:
@smeisler I've been thinking about this one - how familiar are you with the workflow? I'm assuming the topup step happens before Eddy? I also wonder if the outputs from this could be used by DRBUDDI. I believe @eurotomania has used Synb0 with DRBUDDI with good results. 

There may have to be two different workflows for synb0 disco depending on whether it's run with Eddy or SHORELine

cookpa:
I definitely want to use this feature, happy to test things

smeisler:
@mattcieslak The idea (@schillkg correct me if I am wrong) is that we can use Synb0 **only** to make a synthesized RPE b0-like image that would then be tunneled into whatever SDC workflow is requested. While the dockerized synb0-DISCO workflow may go through all of the SDC steps, the DIPY version should be a bit more modular, enabling us to do only the synthesis step.

smeisler:
And here is a relevant part from [the paper](https://www.sciencedirect.com/science/article/pii/S0730725X18306179?via%3Dihub#f0005):

> Distortion correction was performed by first applying the trained network to a given MPRAGE dataset, resulting in the synthesized b0 image. The synthesized b0 was rigidly registered to the real b0 image (note that these will not, and do not have to, align perfectly as both motion and susceptibility distortions are estimated with TOPUP), and concatenated, which formed the input to TOPUP. TOPUP was then run on the merged b0 volume. Importantly, when setting up the acquisition parameters, the readout time (i.e., time between the center of the first echo and center of last echo) for the synthesized image is set to 0 (while the real b0 retains the correct readout time). This tells the algorithm that the synthesized b0 has an infinite bandwidth in the PE direction, with no distortions, thus fixing its geometry when estimating the susceptibility field. **After this**, conventional processing is performed, including [eddy current](https://www.sciencedirect.com/topics/physics-and-astronomy/eddy-current) correction (FSL eddy) and non-linear [least squares](https://www.sciencedirect.com/topics/medicine-and-dentistry/least-square-analysis) diffusion tensor fitting and fractional anisotropy (FA) calculation.

DelinteNicolas:
Any update on this pull request ? This feature would be a great addition

smeisler:
Hi @emilyharriott,

The new synb0 PR, which we are still working on (got a lot of distractions along the way :/ ) is at #905.

In the meantime, what I would do is run synB0 using an average (or good-quality representative) b0, save the `b0_u.nii.gz` as an fmap epi in BIDS format, then make a JSON that has the reverse phase encoded direction and right `IntendedFor` key. I would set the `TotalReadoutTime` to be something like 0.05. I think that's the only keys you need for the fmap?

emilyharriott:
@smeisler once again this is beyond helpful - thank you so much! 

what if my average/good quality representative b0 (that went into synb0 disco to create b0_u.nii.gz) was in the i direction and my dw image (that will go into qsiprep) is in the j direction? in the .json file I'm making, do I list the direction as j or i or even j-?

smeisler:
hi @emilyharriott the b0 from the original DWI will always be same phase encoding direction as the DWI. You assign a new phase encoding direction (different sign, same axis) to the synb0 output so it can be used in SDC.

Issue #482, opened 2022-10-24
FSL dependency


mb3152:
I would like to remove the FSL dependency from QSIPrep.

mattcieslak:
Can you open up a PR and enable contributions from other editors? I think with TORTOISE included we can remove it without loss of performance

mb3152:
Of course, should I do that when we are done over here?

mb3152:
Got it, @jbh1091 is working on it for now.

jbh1091:
@mattcieslak I'm looking at FAST usage right now. I basically see 2 options for replacement: Freesurfer segmentation and AFNI's 3dSeg. Both have differences that prevent a quick plug-and-play swap (no CSF mask in Freesurfer, no TPMs in AFNI). Did you have a solution in mind?

@mb3152 mentioned FAST outputs are already not used much, so deleting the current usage and added segmentation references later in the pipeline when needed is also an option.

mattcieslak:
We can remove FAST from preprocessing entirely. Nothing in the preprocessing workflow uses a brain segmentation other than distortion correction plotting. Maybe we could use Atropos from ANTs, or 3dSeg sounds interesting too.

The trickiest thing to replace is going to be `avscale`, which we use to convert itk transforms to FSL transforms and from those calculate the head motion parameters. 

I have no idea how FSL calculates these motion parameters. They're different from what you get if you read the itk transform files as Euler3DTransforms and get the translations and rotations.

jbh1091:
Totally separate question: I know TORTOISE will be replacing eddy and topup; does that replacement extend to FUGUE and PRELUDE as well? From what I've been reading they do similar things (EPI distortion correction) just with a different method.

mattcieslak:
These also require different fieldmap acquisitions. AFAIK there is nothing out there that does exactly what fugue and prelude do, but unless you acquire GRE fieldmaps (not as popular nowadays) these programs are not useful

jbh1091:
OK that makes sense, thanks. If there isn't a good replacement but it's also not used much, I'll just plan to remove them entirely.

mb3152:
@mattcieslak can we use Matthew Brett's code for the `avscale`?

https://matthew-brett.github.io/transforms3d/reference/transforms3d.affines.html

mattcieslak:
If you can show that it finds the same values as avscale then great, but I've not found anything (including ITK) that produces the same values as avscale

mb3152:
[...]

PR #548, opened 2023-04-17
Replace fsl split

jbh1091:
Part of a series of branches and PRs addressing issue #482.
 
Replaces FSL's Split function with AFNI's 3dTsplit (for the DWI files) and custom code (for bvals and bvecs). Outputs and inputs should match the previous version exactly. FSL Split is still used in workflows that otherwise require FSL (e.g. eddy).


mattcieslak:
Thank you!! Did you get a chance to try this on local data to see if the quartering artifact pops up? One of the places I suspected might have introduced this is if TSplit changes the header in the split images

PR #557, opened 2023-04-25
[FIX] remove pre bids-filter acq type argument

octomike:
I'm debugging an issue and found that this option is present in the cli code, but never used anywhere. This was added in  d08a171c46555446e56bb9eb142f035ec37f8451  and I **assume** it was made obsolete with the `bids-filter-file` argument?

I was about to use this option when I figured out it would have made no difference anyway and I want to save others from falling into this little trap.


mattcieslak:
@octomike thanks for bringing this up. There are too many arguments to qsiprep anyways, happy to lose one that didn't do anything

PR #571, opened 2023-05-12
pyAFQ pipeline: Use previous analysis derivatives as inputs for masks: seed and brain.

### LINKED ISSUE #563: pyAFQ pipeline uses 0 FA as seed threshold?
Seed threshold is set to 0 here: 

https://github.com/PennLINC/qsiprep/blob/master/qsiprep/data/pipelines/pyafq_tractometry.json#L20

Does that mean that pyAFQ uses FA=0 as default threshold for the seed mask? It definitely looks like it from data we've looked at.

Or does it propagate a binary WM mask from the previous qsiprep steps? Is there some way to make that the default behavior?

arokem:

This changes the defaults for seed mask and brain mask for the pyAFQ pipeline, using pyAFQ's mask definition API. 

Might resolve #563


36000:
The same change should be made to this file: https://github.com/PennLINC/qsiprep/blob/master/qsiprep/data/pipelines/mrtrix_multishell_msmt_pyafq_tractometry.json

qsiprep/data/pipelines/pyafq_tractometry.json
```diff
@@ -16,7 +16,7 @@
                 "directions": "prob",
                 "max_angle": 30.0,
                 "sphere": "",
-                "seed_mask": "",
+                "seed_mask": "ImageFile('suffix': 'WM', {'scope': 'qsiprep', 'space': None})",
```
arokem, line 19: Hey @36000 : should this use an `ImageFile` or a `MaskFile` object? Would this work as written here? That is, do these names (`ImageFile`, `MaskFile`) exist in the namespace in which this config is run?

arokem: Also, can anyone tell me how mrtrix pipelines choose where to place seeds? In particular, the msmt_noACT pipeline?

qsiprep/workflows/recon/pyafq.py
```diff
@@ -79,6 +79,7 @@ def init_pyafq_wf(omp_nthreads, available_anatomical_data,
             ('bval_file', 'bval_file'),
             ('bvec_file', 'bvec_file'),
             ('dwi_mask', 'mask_file'),
+            ('wm_mask', 'wm_mask_file'),
```
arokem, line 82: @mattcieslak : what should we put here for `wm_mask`?

arokem:
Updated (with help from @36000, thank you!) to do this the right way. The config is now explicit about using the `WM_prob` output to avoid confusion (it turns out that I was wrong in #563, because the default settings use FA with 0.2 as a threshold, but that logic is a bit hidden in pyAFQ, which we will also fix separately...), and a probability of 0.5 as the threshold.

tsalo:
I've opened https://github.com/PennLINC/qsirecon/issues/170 to track this in QSIRecon. I'll close this PR since there's no way to merge it in QSIPrep but I have a link to it from the QSIRecon issue so someone can copy over the changes to their own branch at some point.

Issue #612, opened 2023-08-11
`qsiprep` says BIDS dir does not exist


jhlegarreta:
I would like to use `qsiprep` to preprocess some dMRI data.

I am using a sandbox-built Singularity image (as building the image failed otherwise), and I am calling `qsiprep` using the following script:

```
#!/bin/bash

fs_license_path=/usr/local/freesurfer/license.txt

singularity_image_path=/mnt/data/containers/qsiprep/qsiprep-0.19.0.sif
in_bids_path=/mnt/data/my/path/
participant="0001"
out_path=/mnt/data/test_qsiprep/

recon_input=$bids_path
output_resolution=1

work_dir=/mnt/data/workdir/

singularity run --cleanenv \
  --bind $in_bids_path:/data:ro \
  --bind $out_path:/out \
  --bind $fs_license_path:/opt/freesurfer/license.txt \
  --bind $work_dir \
  $singularity_image_path \
  $in_bids_path \
  $out_path \
  participant \
  --output-resolution $output_resolution \
  --fs-license-file /opt/freesurfer/license.txt \
  --work_dir $work_dir \
```
The script yields:

```
Making sure the input data is BIDS compliant (warnings can be ignored in most cases).
/mnt/data/my/path/ does not exist
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/qsiprep", line 8, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/cli/run.py", line 607, in main
    validate_bids(opts)
  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/cli/run.py", line 552, in validate_bids
    validate_input_dir(os.name, opts.bids_dir, opts.participant_label)
  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/utils/bids.py", line 299, in validate_input_dir
    subprocess.check_call(['bids-validator', bids_dir, '-c', temp.name])
  File "/usr/local/miniconda/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['bids-validator', PosixPath('/mnt/data/my/path/'), '-c', '/tmp/tmpg23aj4lu']' returned non-zero exit status 2.
```

I know I can bypass the BIDS compliance step, but I want to check it. A few notes:
- I have previously validated the BIDS directory with the JS BIDS validator, with a successful result:
```
[...]

PR #621, opened 2023-08-22
DOC: Remove the destination paths in container documentation

jhlegarreta:
Remove the destination paths for the input and output binding input arguments in the `Docker` and `Singularity` container usage instructions.

Avoids, for example, errors that mean that no (BIDS) data was effectively found, e.g.:
```
Process Process-2:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/cli/run.py", line 833, in build_qsiprep_workflow
    layout = BIDSLayout(
  File "/usr/local/miniconda/lib/python3.8/site-packages/bids/layout/layout.py", line 135, in __init__
    root, description = validate_root(root, validate)
  File "/usr/local/miniconda/lib/python3.8/site-packages/bids/layout/validation.py", line 70, in validate_root
    raise ValueError("BIDS root does not exist: %s" % root)
ValueError: BIDS root does not exist: /mnt/data/my/path/
```

Related to issue #612.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -21,38 +21,44 @@ Docker Container
 In order to run qsiprep in a Docker container, Docker must be `installed
 <https://docs.docker.com/engine/installation/>`_.
 Once Docker is installed, the recommended way to run qsiprep is to use the
-``qsiprep-docker`` wrapper, which requires Python and an Internet connection.
+``qsiprep-docker`` wrapper, which requires Python and an Internet connection
+and that you install the ``qsiprep-container`` package with ``pip``.
 
-.. note:: If running Docker Desktop on MacOS, be sure to set the memory to 6 or more GB.
-    Too little memory assigned to Docker Desktop can result in a message like ``Killed.``
+.. note:: If running Docker Desktop on MacOS (or via Docker Desktop), be sure to set
+    the memory to 6 or more GB. Too little memory assigned to Docker Desktop can result
+    in a message like ``Killed.``
 
 When run, ``qsiprep-docker`` will generate a Docker command line for you,
 print it out for reporting purposes, and then run the command, e.g.::
 
     $ qsiprep-docker /path/to/data/dir /path/to/output/dir participant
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@mattcieslak I found other appearances of the destination paths in:
https://github.com/PennLINC/qsiprep/blob/master/docs/contributors.rst#contributing-to-qsiprep

They also point to very specific destinations paths (and Python 3.6, which reached its EOL on 2021-12-23). Not sure whether I am missing something about such specific destination paths, or whether they should be removed. Let me know what you think.

mattcieslak:
3.6 is too old, I believe at least 3.8 is required not. So that is incorrect in the current documentation

jhlegarreta:
> 3.6 is too old, I believe at least 3.8 is required not. So that is incorrect in the current documentation

I know Matt. I do not have the bandwidth to check whether a simple change from 3.6 to 3.8 would make things work as they are written; if somebody in your lab has the time, I would be happy to submit a separate PR to fix the documentation.

Otherwise, would removing the destination paths altogether be OK?

What about the read only flags? Can you test them or do you want them to stay?

mattcieslak:
Added a few edits, wdyt?

jhlegarreta:
They look good Matt, but I do not have that much experience with `docker` so as to spot inaccurate instructions. Also, I have not tested them.

There are still some references to Python 3.6 in https://github.com/PennLINC/qsiprep/blob/master/docs/contributors.rst#contributing-to-qsiprep, as highlighted in https://github.com/PennLINC/qsiprep/pull/621#issuecomment-1688898659.

Thanks.

mattcieslak:
I will edit the other outdated info in a new PR. Thanks again!

PR #626, opened 2023-08-26
DOC: Improve installation instructions

jhlegarreta:
Improve installation instructions:
- Explicitly mention the possibility of installing `qsiprep` through PyPI, and the required additional dependencies.
- Clarify how this is different from using containers and the `qsiprep-container` PyPI package.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -4,14 +4,25 @@
 Installation
 ------------
 
-There are two easy ways to use qsiprep:
-in a `Docker Container`_, or in a `Singularity Container`_.
-Using a local container method is highly recommended.
-Once you are ready to run qsiprep, see Usage_ for details.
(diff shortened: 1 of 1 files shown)


tsalo:
I believe we removed the PyPi deployment. @mattcieslak is that correct?

tsalo:
Okay cool. @jhlegarreta I can take over this PR since it's been so long since you opened it (sorry about that), unless you have the time to work on it now?

jhlegarreta:
@tsalo I see that you have already opened a PR and I see that the text is different from what I proposed, discouraging the use of `PyPI`, so I guess you have a better understanding of who the tool should be installed. Thanks.

tsalo:
Sorry, I got a little over-eager. I think with any of the preps, installing with PyPi is strongly discouraged except for power users (who are probably forking the repo and cloning it locally TBH).

PR #635, opened 2023-09-11
Update pyAFQ version to 1.1

arokem:
Would this build a new qsiprep container with the new pyAFQ version? Would a user then have to use `latest` to access this version of pyAFQ, until a new RC of qsiprep is up?

### DIFF
--- a/qsiprep/cli/run.py
+++ b/qsiprep/cli/run.py
@@ -840,7 +840,7 @@ def build_qsiprep_workflow(opts, retval):
                     "stimuli",
                     "sourcedata",
                     "models",
-                    re.compile(r"^\.")))
+                    re.compile(r"\/\.\w+|^\.\w+")))
 
     subject_list = collect_participants(
         layout, participant_label=opts.participant_label)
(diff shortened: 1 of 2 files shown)


arokem:
Oh - this is going to be an issue for this upgrade: 

https://app.circleci.com/pipelines/github/PennLINC/qsiprep/1523/workflows/f315e32e-057a-49e9-89f7-b3a2b521701e/jobs/15292

```
ERROR: Cannot install qsiprep and qsiprep==0.19.0+4.gfc32817 because these package versions have conflicting dependencies.

The conflict is caused by:
    qsiprep 0.19.0+4.gfc32817 depends on pybids<0.16.1
    pyafq 1.1 depends on pybids>=0.16.2

```

How strict is the need for the earlier pybids here?

mattcieslak:
not strict at all, we can bump pybids. Once this passes I'll do a RC!

mattcieslak:
Turns out that the new pybids needs a different regex to filter out dotfiles. Credit goes to @tsalo for this discovery

mattcieslak:
We ended up adding an even newer pyafq in #729

PR #638, opened 2023-09-22
DOC: Clarify the output space convention

jhlegarreta:
Clarify the output space convention: make it clear that all data is according to the AC-PC convention, and describe the changes this implies to the data.


mattcieslak (approved):

mattcieslak:
Thanks again!

PR #652, opened 2023-10-26
Fix RTD build

tsalo:

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,6 +3,14 @@ version: 2
 conda:
   environment: .readthedocs/environment.yaml
 
+build:
+  os: ubuntu-22.04
+  tools:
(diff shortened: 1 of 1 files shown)


tsalo:
Current failure seems to be due to https://github.com/daducci/AMICO/issues/164.

mattcieslak:
is there a section in the readthedocs config where we can apt-get install blas and lapack? or maybe conda install them?

tsalo:
I think maybe we can install them with `apt_packages`. I'll give that a try.

https://docs.readthedocs.io/en/stable/config-file/v2.html#conda

tsalo:
YAY!!!!!

tsalo:
I did manage to mess up the build for pypi. I'll need to figure out how it works with the new packaging.

PR #678, opened 2024-01-24
Apply stylistic changes to `workflows/base.py`

tsalo:
Closes none, but relates to the hackathon we're planning.

- Add `# fmt:skip` comments to the ends of `workflow.connect` calls, which will make any autoformatters and linters skip those calls. This will keep black (or whatever autoformatter we want) from messing up workflow connections, which are typically pretty ugly post-auto-formatting.
- Run black and isort on the file.
- Make minor changes to the docstrings.
- Remove an unused import.
- Remove unused parameters from `init_single_subject_wf`: skull_strip_template, skull_strip_fixed_seed, freesurfer, hires, prefer_dedicated_fmaps, and use_syn.
    - I also removed the parameters from `init_qsiprep_wf` and added `(IGNORED)` to the descriptions of the related command line arguments.

### DIFF
--- a/tests/smoke_tests.py
+++ b/tests/smoke_tests.py
@@ -6,173 +6,160 @@ from qsiprep.workflows.base import init_qsiprep_wf
 @pytest.mark.smoke
 def test_preproc_pepolar_sdc(tmp_path):
     # Get the empty bids data
-    bids_root = pkgrf('qsiprep', 'data/abcd')
+    bids_root = pkgrf("qsiprep", "data/abcd")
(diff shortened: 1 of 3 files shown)


tsalo:
@mattcieslak if it's okay, I'd love to remove the unused parameters to `init_single_subject_wf`. WDYT?

mattcieslak:
I agree completely!

tests/smoke_tests.py
mattcieslak, line 1: This file is never actually used. It's left over from when I briefly tried to use pytests on the workflows. This is another direction that should be revisited

mattcieslak (approved):

PR #679, opened 2024-01-24
Support complex-valued dwidenoising

### LINKED ISSUE #677: Leverage phase data, when available, at denoising stage
`dwidenoise` (MRTRIX3 >= 3.0.0) can handle complex-valued data, which improves the denoising.

Here's what the documentation says:

>Note that dwidenoise does not correct for non-Gaussian noise biases present in magnitude-reconstructed MRI images. Including MRI phase images can reduce such Rician bias [[CorderoGrande2019]](https://mrtrix.readthedocs.io/en/latest/reference/references.html#corderogrande2019), and the command now supports complex input data. To this end, users can run:
>
> ```
> mrcalc dwi_magnitude.mif dwi_phase_rad.mif -polar dwi_complex.mif
> dwidenoise dwi_complex.mif out.mif -noise noise.mif
> mrcalc out.mif -abs out_magnitude.mif
> ```
>
> Where dwi_magnitude.mif and dwi_phase_rad.mif are the magnitude and phase images respectively. Note that the code above assumes that the phase is scaled in radians.
>
>Whilst using complex data can effectively reduce Rician noise bias, dwidenoise currently does not account for spatial noise correlations introduced due to in-plane accelleration. [[CorderoGrande2019]](https://mrtrix.readthedocs.io/en/latest/reference/references.html#corderogrande2019) addresses these additional effects at reconstruction, but this is not implemented in MRtrix3.

On QSIPrep's side, I think it would need to (1) collect the phase data when available, (2) create the complex-valued files from the magnitude and phase data, (3) run `dwidenoise` on the complex files, and (4) split out the magnitude data from the denoised complex files.

tsalo:
Closes #677.

We can probably use [ds004666](https://openneuro.org/datasets/ds004666) for testing/validation, though we'll need to add sidecar files with some metadata (PhaseEncodingDirection and TotalReadoutTime).

- In `collect_data`, select "part-mag" or no part entity for "dwi" data.
    - This will ensure that regular DWI processing doesn't accidentally collect phase files.
- Add steps to perform complex-valued dwidenoise when possible. 
    - Limited to when (1) `--denoise-method` is set to `dwidenoise`, (2) `--ignore` doesn't include `phase`, (3) `--denoise-after-combining` is not enabled, and (4) phase data are available.
- Add `phase` option to `--ignore` to explicitly disable complex-valued denoising when phase data are available.
- In `init_merge_and_denoise_wf`, use `layout.get_metadata()` to get each DWI file's metadata.


qsiprep/workflows/dwi/merge.py
```diff
@@ -217,188 +252,358 @@ def init_merge_and_denoise_wf(raw_dwi_files,
...
+            DerivativesDataSink(
+                suffix=f"{name}_denoising",
+                source_file=source_file,
+            ),
+            name=f"ds_report_{name}_denoising",
+            run_without_submitting=True,
+            mem_gb=DEFAULT_MEMORY_MIN_GB,
+        )
+
+        if (denoise_method == "dwidenoise") and phase_available:
```
tsalo, line 470: Here is the main change.

tsalo:
It seems like the "combining" step (`not denoise_before_combining`) involves processing (`b0_threshold` and `not no_b0_harmonization`) that would mess up the relationship between the magnitude and phase data, so we might need to limit complex dwidenoise to workflows with denoising before combining.

I also don't know what `merging_distortion_groups` really does. Does it just create a cross-run template, or does it do some initial processing to the DWI data _before_ the denoising workflow?

tsalo:
So what I ended up doing is passing along the `BIDSLayout` to `init_merge_and_denoise_wf`. Now it will try to find a phase file for each individual DWI file as it loops through the DWI files. Complex-valued denoising will only be allowed if `denoise_before_combining` is True, since I'm not sure if the extra steps from combining (e.g., the B0 harmonization) would mess up the relationship between the magnitude and phase data. Also not sure that we can assume that complementary magnitude runs will all have phase data.

tsalo:
Tested on the 0p9mm session of ds004666, and it failed on `MergeDWIs`:

```
ValueError: MergeDWIs requires a value for input 'bval_files'. For a list of required inputs, see MergeDWIs.help()
```

qsiprep/workflows/dwi/merge.py
```diff
@@ -167,21 +181,43 @@ def init_merge_and_denoise_wf(
...
             _, fname, _ = split_filename(dwi_file)
             wf_name = _get_wf_name(fname).replace("preproc", "denoise")
+
+            # Try to find associated phase file.
+            # BIDSLayout.get_nearest() doesn't work well for this.
+            ents = layout.get_file(dwi_file).get_entities()
+            ents["part"] = "phase"
+            phase_files = layout.get(**ents)
+            phase_available = False
+            if len(phase_files) == 1:
```
mattcieslak, line 206: If the user hasn't validated their dataset is it possible they'll have more than 1?

qsiprep/workflows/dwi/merge.py
```diff
@@ -277,22 +329,26 @@ def init_merge_and_denoise_wf(
         mem_gb=mem_gb,
         omp_nthreads=omp_nthreads,
         source_file=source_file,
+        use_phase=False,  # can't use phase with concatenated data
```
mattcieslak, line 332: Is this true?

qsiprep/workflows/dwi/merge.py
```diff
@@ -329,14 +454,22 @@ def init_dwi_denoising_wf(
...
+            name=f"buffer{num_buffers:02}",
+        )
 
     buffernodes.append(get_buffernode())
+
     workflow.connect([
-        (inputnode, buffernodes[-1], [('dwi_file', 'dwi_file')]),
+        # The first buffernode is the raw file
+        (inputnode, buffernodes[0], [('dwi_file', 'dwi_file')]),
+        # XXX: Why pass the bval and bvec files through unmodified?
```
mattcieslak, line 467: The bval and bvec won't be changed by denoising

qsiprep/workflows/dwi/merge.py
```diff
@@ -350,7 +483,66 @@ def get_buffernode():
     # Add the steps
     step_num = 1  # Merge inputs start at 1
     if do_denoise:
-        if denoise_method == "dwidenoise":
+        # Add buffernode for denoised DWI
+        buffernodes.append(get_buffernode())
+
+        ds_report_denoising = pe.Node(
```
mattcieslak, line 489: Added issue #707 to revisit soon

qsiprep/interfaces/dwi_merge.py
```diff
@@ -714,3 +715,77 @@ def create_provenance_dataframe(
...
+
+    Notes
+    -----
+    The code is derived from
+    https://github.com/nipreps/sdcflows/blob/c6cd42944f4b6d638716ce020ffe51010e9eb58a/\
+    sdcflows/utils/phasemanip.py#L26.
+
+    License
+    -------
+    Copyright 2021 The NiPreps Developers <nipreps@gmail.com>
```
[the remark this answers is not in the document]
tsalo: Since it only applies to a specific function, I _think_ it only needs to go in the docstring. I took a look at https://www.nipreps.org/community/licensing/#example-of-our-expectations and will push some more comprehensive changes to the docstrings in a second.

qsiprep/interfaces/mrtrix.py
```diff
@@ -1365,3 +1365,54 @@ class TransformHeader(CommandLine):
...
+        position=-1,
+        argstr="-polar %s",
+    )
+
+
+class _PolarToComplexOutputSpec(TraitedSpec):
+    out_file = File(exists=True)
+
+
+class PolarToComplex(CommandLine):
```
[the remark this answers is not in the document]
tsalo: I got it from the docstring of mrcalc: https://mrtrix.readthedocs.io/en/latest/reference/commands/mrcalc.html#complex-numbers

qsiprep/workflows/dwi/merge.py
tsalo: I suppose it's possible if there are other DWI scans that have a superset of this scan's entities. Something like:

```
sub-01_part-mag_dwi.nii.gz
sub-01_part-phase_dwi.nii.gz
sub-01_run-01_part-mag_dwi.nii.gz
sub-01_run-01_part-phase_dwi.nii.gz
```

In that case, `layout.get` using `sub-01_part-mag_dwi.nii.gz` as the base might find both `sub-01_part-phase_dwi.nii.gz` and `sub-01_run-01_part-phase_dwi.nii.gz`. Not sure if that's a valid BIDS dataset, but that situation could happen.

I can make the `get` call more strict by explicitly setting all other entities to `Query.NONE`.

qsiprep/workflows/dwi/merge.py
tsalo: Now that I think about it, probably not. I do think you should keep it disabled until you (or someone else) can do some testing to make sure it's okay.

qsiprep/workflows/dwi/merge.py
tsalo: It's hard for me to tell, but it looks like run-wise bval and bvec files get concatenated in `init_merge_and_denoise_wf` only if `denoise_before_combining` is True. It just seems like this could be made clearer through comments.

qsiprep/workflows/dwi/merge.py
tsalo: Take a look at the version I just pushed. WDYT?

mattcieslak: this should be sufficiently strict!

qsiprep/workflows/dwi/merge.py
mattcieslak: I personally am not a fan of combining before denoising, and it looks like less than 1% of all recorded qsiprep runs have chosen that route. We can leave it false for now but make an issue for the roadmap. I wonder if the scaling of the phase data needs to incorporate all the concatenated runs?

qsiprep/workflows/dwi/merge.py
mattcieslak: agreed. This whole workflow needs comments

mattcieslak (approved):

Issue #184, opened 2020-09-24
Utilising s2v parameters in eddy_config.json


sidchop:
Hi,

I am trying to apply slice-to-volume correction using eddy. However, the [required attributes](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/eddy/UsersGuide#A--s2v_niter) are not considered appropriate within the eddy_params.json file. 

My eddy_params.json file:

`{
  "flm": "linear",
  "slm": "linear",
  "fep": false,
  "interp": "spline",
  "nvoxhp": 1000,
  "fudge_factor": 10,
  "dont_sep_offs_move": false,
  "dont_peas": false,
  "niter": 5,
  "method": "jac",
  "repol": true,
  "num_threads": 1,
  "is_shelled": true,
  "use_cuda": true,
  "cnr_maps": true,
  "residuals": false,
  "output_type": "NIFTI_GZ",
  "args": "",
  "mporder": 20,
  "s2v_niter": 10,
  "s2v_lambda": 5,
  "s2v_interp": "trilinear",
  "slspec": "data/slspec.txt"
}`

and I am calling qsiprep using:
`singularity run --cleanenv --bind data:/data qsiprep-0.11.0.sif \
--participant-label sub-003 \
--use-syn-sdc \
--force-syn \
--output-resolution 2 \
--skip_bids_validation \
--hmc_model eddy \
--eddy-config data/eddy_params.json \
--fs-license-file data/license.txt \
-w data/qsiprep_work \
data/rawdata \
data/derivatives \
participant`

The error is:
`Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.7/site-packages/qsiprep/cli/run.py", line 908, in build_qsiprep_workflow
    force_syn=opts.force_syn
  File "/usr/local/miniconda/lib/python3.7/site-packages/qsiprep/workflows/base.py", line 261, in init_qsiprep_wf
    force_syn=force_syn)
  File "/usr/local/miniconda/lib/python3.7/site-packages/qsiprep/workflows/base.py", line 692, in init_single_subject_wf
    source_file=source_file
[...]

PR #710, opened 2024-03-04
[ENH] Make Eddy's slice2vol much easier to use

mattcieslak:
# Background

Eddy-CUDA has a slice-to-volume motion correction method. It is enabled by the `--mporder <int>` flag and requires that slice times and multiband acceleration factor are defined.

Currently users are accessing this feature by creating their own slice-timing file and mounting it into the container, along with specifying the other options in the eddy config json file (eg #184). This is a lot of extra work and not robust if there are different slice timings in different scans.

# Proposed change

The trickiest part here is figuring out what the slice timings should be if there are multiple scans being concatenated before they're sent to Eddy. If the slice timings are too different, qsiprep should detect it early and exit. We discussed that a 10ms difference in a slice's timing across scans is acceptable. 

I propose we just take the slice timings from the first scan as the official slice timings sent to Eddy.

# To Do

 - [X] Create a json file from the MergeDWI interface that copies slice timings from a real sidecar
 - [X] Pass the sidecar to Eddy via `--json`, along with `--multiband_factor` when `mporder` is in the eddy config
~~- [ ] Perform a sanity check that the slice timings are similar enough that merging them makes sense~~

Moved to #712

### DIFF
--- a/qsiprep/interfaces/dwi_merge.py
+++ b/qsiprep/interfaces/dwi_merge.py
@@ -88,7 +88,7 @@ class MergeDWIs(SimpleInterface):
             )
             merged_metadata_file = op.join(runtime.cwd, "merged_metadata.json")
             with open(merged_metadata_file, "w") as merged_meta_f:
-                json.dump(combined_metadata, merged_meta_f, indent=4)
+                json.dump(combined_metadata, merged_meta_f, sort_keys=True, indent=4)
             self._results["merged_metadata"] = merged_metadata_file
 
         # Get basic qc / provenance per volume
@@ -163,14 +163,31 @@ class MergeDWIs(SimpleInterface):
         return runtime
 
 
-def combine_metadata(scan_list, metadata_dict):
+def combine_metadata(scan_list, metadata_dict, merge_method="first"):
     """Create a merged metadata dictionary.
 
-    Most importantly, combine the slice timings in some way. Here the slice
-    metadata is taken from the first scan.
+    Most importantly, combine the slice timings in some way.
+
+    Parameters
+    ----------
+    scan_list: list
+        List of BIDS inputs in the order in which they'll be concatenated
+    medadata_dict: dict
+        Mapping keys (values in ``scan_list``) to BIDS metadata dictionaries
+    merge_method: str
+        How to combine the metadata when multiple scans are being concatenated.
+        If "first" the metadata from the first scan is selected. Someday other
+        methods like "average" may be added.
+
+    Returns
+    -------
+    metadata: dict
+        A BIDS metadata dictionary
 
     """
-    return metadata_dict[scan_list[0]]
+    if merge_method == "first":
+        return metadata_dict[scan_list[0]]
+    raise NotImplementedError(f"merge_method '{merge_method}' is not implemented")
 
 
 class AveragePEPairsInputSpec(MergeDWIsInputSpec):
--- a/qsiprep/interfaces/dwi_merge.py
+++ b/qsiprep/interfaces/dwi_merge.py
@@ -48,6 +48,7 @@ class MergeDWIsInputSpec(BaseInterfaceInputSpec):
     carpetplot_data = InputMultiObject(
         File(exists=True), mandatory=False, desc="list of carpetplot_data files"
     )
+    scan_metadata = traits.Dict(desc="Dict of metadata for the to-be-combined scans")
 
 
 class MergeDWIsOutputSpec(TraitedSpec):
@@ -55,7 +56,7 @@ class MergeDWIsOutputSpec(TraitedSpec):
     out_bval = File(desc="the merged bval file")
     out_bvec = File(desc="the merged bvec file")
     original_images = traits.List()
-    merged_metadata = traits.Dict()
+    merged_metadata = File(exists=True)
     merged_denoising_confounds = File(exists=True)
     merged_b0_ref = File(exists=True)
     merged_raw_dwi = File(exists=True, mandatory=False)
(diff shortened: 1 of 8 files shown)


qsiprep/interfaces/dwi_merge.py
```diff
@@ -151,6 +163,16 @@ def _run_interface(self, runtime):
...
 
 
+def combine_metadata(scan_list, metadata_dict):
+    """Create a merged metadata dictionary.
+
+    Most importantly, combine the slice timings in some way. Here the slice
+    metadata is taken from the first scan.
+
+    """
+    return metadata_dict[scan_list[0]]
```
tsalo, line 173: What about raising a warning if `SliceTiming` doesn't match across scans? That way, at least users will be aware of an issue.

qsiprep/interfaces/dwi_merge.py
```diff
@@ -151,6 +163,16 @@ def _run_interface(self, runtime):
         return runtime
 
 
+def combine_metadata(scan_list, metadata_dict):
```
tsalo, line 166: Given that you're not actually combining the metadata, maybe rename this to `select_multirun_metadata` or something similar?

qsiprep/interfaces/eddy.py
```diff
@@ -125,6 +130,21 @@ def _run_interface(self, runtime):
...
+            # json file is only allowed if mporder is defined
+            if "mporder" in eddy_config:
+                self._results["json_file"] = self.inputs.json_file
+
+        #  Get the multiband factor from the json file and convert it to an arg
+        if isdefined(self.inputs.json_file):
+            with open(self.inputs.json_file) as json_f:
+                input_sidecar = json.load(json_f)
+            if "MultibandAccelerationFactor" in input_sidecar:
+                self._results["multiband_factor"] = input_sidecar["MultibandAccelerationFactor"]
```
tsalo, line 146: What does Eddy use `multiband_factor` for?

tsalo (commented):
I mostly have thoughts on documentation and function naming. I have a few questions as well.

qsiprep/interfaces/eddy.py
tsalo: Asking because I worry that multiband factor might be used to infer slice timing. There are some cases (e.g., the CMRR EPI sequence with certain combinations of MB factor and number of slices) where slices won't actually be in the typical order.

cookpa: The only use I can think of for it would be for qsiprep to check that mporder is < MultibandAccelerationFactor

qsiprep/interfaces/dwi_merge.py
mattcieslak: That's still in my TODOs for this PR. I was going to have that somewhere in utils/grouping.py so there is an early error

qsiprep/interfaces/dwi_merge.py
mattcieslak: I was thinking about adding an option in the future where the slice timings could be averaged or the median taken per slice across the combined scans. I was hoping to get some feedback on that though before implementing

qsiprep/interfaces/eddy.py
mattcieslak: It uses it to determine how many groups of slices it should expect. I think it is used internally to help figure out how to divide up the slice timings if it receives a json file instead of its "slspec" format

cookpa: Oh wait, this is going into the JSON file to pass to eddy? Yes I think it needs the factor to know how many groups there are

cookpa: I think it doesn't infer slice timings from the number of groups unless you use the deprecated `--mb` and `--mb_offs` options

qsiprep/interfaces/dwi_merge.py
cookpa: I'm not sure how much it matters, because slspec.txt doesn't even contain any timings, just groupings, so I think it just needs the groups to be consistent.

But I can't think of circumstances where the slice timings would be substantially different across runs, unless you're also doing something inadvisable like altering TE.

mattcieslak: In that case I may just stick with using the exact slice times from the first scan. If all the other scans are within 10ms it shouldn't matter too much. @cookpa should I also add a check in the grouping function to make sure TEs are very similar?

qsiprep/interfaces/eddy.py
cookpa: Yes per the user guide "The current version of eddy allows for a more detailed description of the multi-band structure (see --slspec [below](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/eddy/UsersGuide#A--slspec)), and if one wants to do slice-to-volume motion correction it is necessary to use that description. It is still possible to use --mb, but it is discouraged and it will be deprecated in the next release."

qsiprep/interfaces/dwi_merge.py
cookpa: Sorry I think I should have said TR there. But if that varies, slice timings probably would too.

I don't know if you want to get into vetting all the imaging parameters before merging, or if that should be a user responsibility

tsalo: Would it make sense to raise an error if the slice times are very different? Something like CuBIDS should probably give the different runs different `acq` values in that situation.

qsiprep/interfaces/dwi_merge.py
```diff
@@ -151,6 +163,16 @@ def _run_interface(self, runtime):
...
+    """Create a merged metadata dictionary.
+
+    Most importantly, combine the slice timings in some way. Here the slice
+    metadata is taken from the first scan.
+
+    """
```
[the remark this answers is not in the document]
mattcieslak: I may save that for a refactor PR. Added the docstring though

qsiprep/interfaces/eddy.py
mattcieslak: ok, removed this section so mb won't be defined

qsiprep/interfaces/dwi_merge.py
mattcieslak: I'm about half way through this and am thinking it should go in another PR

PR #712, opened 2024-03-05
[ENH] Perform a sanity check on scans that are to be merged

mattcieslak:
The merging done in QSIPrep was written before multipart IDs or b0 field identifiers were a part of BIDS. The strategy is based on the fields used in TOPUP/Eddy to define distortion groups, namely TotalReadoutTime and PhaseEncodingDirection. 

#710 makes slice-to-volume correction possible in Eddy but depends on concatenated scans having compatible slice timings, which is not currently checked for. A few other features of the to-be-concatenated images should also be checked.


mattcieslak:
This was replaced by `qsiplan`

PR #724, opened 2024-04-18
[FIX] Update Everything

mattcieslak:
Due to significant bugs in FSL, and FSL's new conda-based installation system there had to be a major overhaul of the qsiprep_build docker image.

### DIFF
--- a/qsiprep/interfaces/tortoise.py
+++ b/qsiprep/interfaces/tortoise.py
@@ -42,10 +42,7 @@ LOGGER = logging.getLogger("nipype.interface")
 
 SLOPPY_DRBUDDI = (
     "--DRBUDDI_stage "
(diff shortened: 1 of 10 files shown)


qsiprep/interfaces/tortoise.py
```diff
@@ -219,6 +204,11 @@ def write_dummy_bmtxt(nii_file):
 
 
 class _DRBUDDIInputSpec(TORTOISEInputSpec):
+    num_threads = traits.Int(
+        desc="number of OMP threads",
+        argstr="--ncores %d",
+        help="Number of cores to use in the CPU version. The default is 50% of system cores.",
```
mattcieslak, line 210:
```suggestion
        help="Number of cores to use in the CPU version. The default is 50% of system cores.",
        nohash=True,
```

PR #737, opened 2024-04-26
[ENH] Allow topup to use multiple threads

### LINKED ISSUE #715: Allow `topup` to use multiple threads
## Summary

`topup` can be started with multiple threads

```shell
❯ topup --help 2>&1| grep threads
	--nthr		Number of threads to use (cannot be greater than numbers of hardware cores), default 1
```

Multiple threads can produce decent speedups, but `topup` is currently limited to just 1 thread (around [here](https://github.com/PennLINC/qsiprep/blob/89ba85b3cbdba4c0afad875c29d63154c6aa1fc7/qsiprep/workflows/dwi/fsl.py#L287)). 

## Additional details

[`nipype` doesn't have an option relating to number of threads](https://nipype.readthedocs.io/en/latest/api/generated/nipype.interfaces.fsl.epi.html#topup), and it looks like [the number of threads is not set in the `sdcflows` implementation](https://github.com/nipreps/sdcflows/blob/73d717a23d5ea4672c9f80be520a47fae4bd7b17/sdcflows/workflows/fit/pepolar.py#L37).

## Next steps

To work around the nipype implementation, one could use the `args` option, as in, for example

```python
topup.inputs.args = (f"--nthr {omp_nthreads}")
```

Another option would be to define a new subclass of `nipype.interfaces.fsl.epi.TOPUP` that has a `num_threads` inputspec.

This seems like this would be relatively simple, so I'd be happy to submit a pull request.

psadil:

Set `nthr` flag in `topup` to the value given by user for `omp-nthreads`.

Closes #715

### DIFF
--- a/qsiprep/workflows/dwi/fsl.py
+++ b/qsiprep/workflows/dwi/fsl.py
@@ -284,7 +284,13 @@ def init_fsl_hmc_wf(
             gather_inputs.inputs.epi_fmaps = scan_groups["fieldmap_info"]["epi"]
 
         outputnode.inputs.sdc_method = "TOPUP"
-        topup = pe.Node(fsl.TOPUP(out_field="fieldmap_HZ.nii.gz", scale=1), name="topup")
+        topup = pe.Node(
+            fsl.TOPUP(
+                out_field="fieldmap_HZ.nii.gz", 
+                scale=1, 
+                args=f"--nthr={omp_nthreads}"),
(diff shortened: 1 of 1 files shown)


mattcieslak:
Hi @psadil I ended up merging a nearly identical PR into 0.21.3. Thanks for contributing, and this still qualifies you to be added to the author list if you'd like.

PR #738, opened 2024-04-30
[FIX] PyAFQ was not being written to derivatives

mattcieslak:
Issue is described [here](https://neurostars.org/t/qsi-reconstruction-writes-to-work-dir-but-not-derivatives/29209/1).

The PyAFQ workflows needed a `"qsirecon_suffix"` in their datasink.


dkp:
FYI: As of qsiprep 0.22.1, pyAFQ derivatives are still only being written to to the work directory under qsirecon_0_22_wf.

PR #746, opened 2024-05-09
[DOC] Update reconstruction.rst

chiuhoward:


mattcieslak:
I think we got this info in the new recon docs at qsirecon.readthedocs.io

PR #756, opened 2024-05-31
[FIX]: changed typo and removed native_reference

chiuhoward:
Changing what @richford and I think is a typo - native_template_brain_file to native_template_mask_file to be consistent with the other usage of this variable. We also removed native_reference as we did not see the variable being used elsewhere in the code.

1. Was native_reference a placeholder for WIP? If so, could you point us to where we should use it? We'll be happy to finish it. 
2. Could you also suggest a simple test that we could write for new native templates?


tsalo:
@chiuhoward sorry, I think we changed GetTemplate substantially and the native stuff isn't in there anymore. I think we can close this.

PR #760, opened 2024-06-04
Drop functions to read sidecar JSONs

### LINKED ISSUE #759: Remove `ReadSidecarJSON` and `get_metadata_for_nifti`
These methods use regular expressions to find sidecar JSON files. Now that QSIPrep has access to the BIDSLayout through the config object, there's no need to rely on custom code to find metadata.

tsalo:
Closes #759.

- Remove `qsiprep.interfaces.bids.ReadSidecarJSON` (this was actually unused, as the interface that _was_ used in `init_phdiff_wf` was the one imported from niworkflows).
- Remove `qsiprep.interfaces.bids.get_metadata_for_nifti` in favor of `config.execution.layout.get_metadata`.
- Remove niworkflows's `ReadSidecarJSON` calls in favor of `config.execution.layout.get_metadata`.
- Add `phase_meta` input to `init_phdiff_wf`. This actually matches the SDCFlows workflow a bit better as an added bonus.


tsalo:
I haven't removed `read_nifti_sidecar`, which is a big barrier to using the inheritance principle in DWI datasets.

qsiprep/workflows/dwi/merge.py
```diff
@@ -115,7 +114,9 @@ def init_merge_and_denoise_wf(
             bids_dwi_files=raw_dwi_files,
             b0_threshold=config.workflow.b0_threshold,
             harmonize_b0_intensities=not config.workflow.no_b0_harmonization,
-            scan_metadata={scan: get_metadata_for_nifti(scan) for scan in raw_dwi_files},
+            scan_metadata={
+                scan: config.execution.layout.get_metadata(scan) for scan in raw_dwi_files
```
mattcieslak, line 118: much better!

mattcieslak (approved):

mattcieslak:
The readthedocs error is very strange, I don't think this PR changes anything in extras_require

tsalo:
No, but I'm seeing the same failure in #766. I think it might be an issue with DiPy.

qsiprep/workflows/dwi/merge.py
tsalo: The only issue is that `read_nifti_sidecar` is still used in places. It looks like there are a few interfaces/functions (`get_distortion_grouping` and `get_best_b0_topup_inputs_from`) that use `read_nifti_sidecar` and do some complicated stuff, so I didn't try to remove it.

mattcieslak:
I think this will work, but I'm wondering if we should hold off on major refactors until we add a couple more CI tests (like for GRE fieldmaps). Wdyt?

tsalo:
I'm happy to wait on it.

PR #764, opened 2024-06-14
MAINT: Upgrade pyAFQ to 1.3.2

arokem:


tsalo:
It might be best to wait until #763 is merged (so we can see test outputs without having to run locally) before merging this.

mattcieslak:
Thanks @arokem!

PR #766, opened 2024-06-24
[FIX] Update autotrack workflow to get all dsi studio bundles

mattcieslak:
The DSI Studio update in 0.21 resulted in some bundles not being included in the autotrack results. This should fix the problem.

Thanks to @ameliecr, @hamsiradhakrishnan, @smeisler for reporting and fixing!


mattcieslak:
This PR got out of date very quickly. The fix was simple so I opened a new one (#783) and will close this.

Issue #772, opened 2024-07-08
SIFT2 mu 


araikes:
@mattcieslak 

When using SIFT2, is the `mu.txt` file getting written somewhere and/or are the outputs stored somewhere? I haven't found it. If not, I have some code which works that will write it to the ReconDerivatives.... I just wasn't sure if I was missing it somewhere.

mattcieslak:
solved in #774, thanks again @araikes

PR #774, opened 2024-07-09
Save SIFT2 mu parameter

araikes:
Addresses #772

Currently the SIFT2 mu parameter is connected to mu in the `outputnode`, but is never written to the output folder. Given that this is a necessary correction factor when using sum of streamlines connectomes after SIFT2, this will write the mu parameter to the derivatives folder.


mattcieslak:
Thanks for implementing this @araikes. It looks good, the tests are failing because they weren't expecting the mu files. I added these and hopefully everything will pass now 🤞

PR #780, opened 2024-07-11
WIP [ENH] Add singleshell benchmarking recon workflow

mattcieslak:
cc @ameliecr

We're working on a project to compare tractometry methods using different reconstruction methods on single-shell acquisitions. There are too many initial steps that need to be shared for autotrack to use the simple recon spec json format. So this workflow would break up the initial calculation of qa and iso, and registration into separate steps.

Relevant thread: https://groups.google.com/g/dsi-studio/c/emMz-K8ewlo

### DIFF
--- a/qsiprep/workflows/recon/benchmarking.py
+++ b/qsiprep/workflows/recon/benchmarking.py
@@ -1,16 +1,11 @@
+import logging
 import nipype.pipeline.engine as pe
 from nipype.interfaces import utility as niu
 
 from ... import config
 from ...engine import Workflow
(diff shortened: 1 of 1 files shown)


mattcieslak:
yes, this will be important for the clinical data processing. @smeisler this may be a better PR to get started on than synb0 for now since we'll need it for the clinical dmri. It is supposed to add @ameliecr's work on ss3t-autotrack into qsirecon

tsalo:
I've opened https://github.com/PennLINC/qsirecon/issues/171 for this and linked to this PR.

PR #801, opened 2024-08-08
[DOCS] Update reconstruction.rst to use dashes in flags instead of underscores

arokem:
Looks like that is the correct usage, based on [this page](https://qsiprep.readthedocs.io/en/latest/usage.html).


tsalo:
Since we've undergone The Great Split, I think we can close this PR. I've commented on https://github.com/PennLINC/qsirecon/pull/56 about making the corresponding changes in QSIRecon.

Issue #819, opened 2024-09-13
Upsampling is introducing artifact


mattcieslak:
We got a report that when users upsample data (ie provide an `--output-resolution` that is smaller than the input voxel size) that there are artifacts in derived maps like FA.

Things to change:

 - [x] Change documentation to note the importance of `--output-resolution`
 - [x] Change the interpolation to Linear when upsampling is detected

PR #820, opened 2024-09-13
[FIX] Change to Linear for upsampling

mattcieslak:
Addresses #819

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -17,8 +17,13 @@ One way to process these data would be to call ``qsiprep`` like this::
 
   qsiprep \
     /path/to/inputs /path/to/outputs participant \
-    --output-resolution 1.2 \
+    --output-resolution X \
     --fs-license-file /path/to/license.txt
+
+.. warning::
+   The above example sets the ``--output-resolution`` to ``X``, where in
+   real usage this number should be set to a value in millimeters.
+   See :ref:`output_resolution` for specifics.
 
 
 Grouping scans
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -87,7 +95,7 @@ your output to at least 1.3mm resolution. By choosing this resolution here,
 it means your data will only be interpolated once: head motion correction,
 susceptibility distortion correction, coregistration and upsampling will be
 done in a single step. If your are upsampling your data by more than 10%,
```
tsalo, line 97:
```suggestion
done in a single step. If you are upsampling your data by more than 10%,
```

tsalo:
Why do you use Lanczos windowed sinc for downsampling and upsampling to >90% of the original voxel size. Would it make more sense to just use linear no matter what?

mattcieslak:
> Why do you use Lanczos windowed sinc for downsampling and upsampling to >90% of the original voxel size. Would it make more sense to just use linear no matter what?

Lanczos Sinc is very accurate if certain conditions are met. As long as there aren't sharp cutoffs between high and low intensity and the image resolution in the output isn't higher than the input data it is a great default (and is the default in fmriprep). A added the 90% as some wiggle room for weird voxel sizes.

Linear intepolation ends up smoothing the data, which is not something we want to do unintentionally. I think it makes sense for upsampling, since it won't introduce the ringing effects like a Sinc interpolator will. 

I didn't think BSpline would add ringing (particularly the relatively simple BSpline[3] we were using), but in some cases it looks like can. Although, this may only happen if there is significant upsampling. I'm not sure what the native resolution was of the data that showed the ringing artifacts.

cookpa (approved):
LGTM

tsalo (approved):

PR #846, opened 2024-11-07
Fix small typo in installation instruction

arokem:
"Sigularity" => "Singularity"


tsalo:
@arokem should we change Singularity to Apptainer throughout the documentation?

arokem:
I'll do some more work here to include apptainer and explain that it's sorta the same thing.

arokem:
I've added "Apptainer" in more places. In some places it seems warranted to have both Singularity and Apptainer, just in case (?)

tsalo (approved):
LGTM!

PR #849, opened 2024-11-07
Add ability to ignore phase data

tsalo:
The ability to ignore phase data was added in #679, but it looks like it was accidentally removed in #745. This PR adds it back in.

- Add "phase" option to `--ignore` parameter.

### DIFF
--- a/qsiprep/cli/parser.py
+++ b/qsiprep/cli/parser.py
@@ -302,7 +302,7 @@ def _build_parser(**kwargs):
         action="store",
         nargs="+",
         default=[],
-        choices=["fieldmaps", "sbref", "t2w", "flair", "fmap-jacobian"],
+        choices=["fieldmaps", "sbref", "t2w", "flair", "fmap-jacobian", "phase"],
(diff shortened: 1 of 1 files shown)


lattaai3:
Hello, I tried to fix the issue

This is what I did:

This change adds the 'phase' option to the --ignore parameter in the parser, allowing users to ignore phase data as mentioned in the GitHub issue.

I also created pull request: https://github.com/PennLINC/qsiprep/pull/850


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

mattcieslak (commented):
This looks good to me, do we really not need to change anything anywhere else?

tsalo:
#679 added all of the important parts- we just accidentally dropped the "phase" option for the "--ignore" parameter in #745, so that's the only change we should need.

PR #856, opened 2024-11-08
Infer the appropriate MNIInfant cohort based on the participant's age

### LINKED ISSUE #826: Infer appropriate MNIInfant cohort based on participant age
## Summary

Nibabies already does this (https://github.com/nipreps/nibabies/blob/master/nibabies/utils/bids.py), and it would be great to support it in QSIPrep.

tsalo:
Closes #826.

- Adapt Nibabies' code to figure out the right cohort to use based on a participant's age.
- Make `--infant` require `--subject-anatomical-reference session` since age can't be inferred across sessions.


qsiprep/workflows/anatomical/volume.py
```diff
@@ -1214,7 +1218,7 @@ def init_anat_derivatives_wf() -> Workflow:
             to="T1w",
             mode="image",
             suffix="xfm",
-            **{"from": config.workflow.anatomical_template},
+            **{"from": anatomical_template},
```
tsalo, line 1221: @mattcieslak I think the only output change will be to these transform files. Instead of `from-MNIInfant_to-T1w` and vice versa, it'll be `from-MNIInfant+<cohort>_to-T1w`.

tsalo: That will have implications for QSIRecon though.

mattcieslak: We're updating that part of qsirecon right now anyways, so this should be ok

mattcieslak (approved):

chiuhoward:
Hi @tsalo, must I have a `.tsv` file to specify the ages? (My data is not longitudinal)

I read the rc1 and Nibabies documentation and tried using `—age-months 2` as an argument, which was invalid as expected. Otherwise is there some way I could specify the cohort from the available keys directly (e.g. from `bids.py L719-739`)?  

If constructing a `.tsv` file is the only way, could I confirm that I just need a column with subjects and a second column with `age_months`? Are column headers needed?

Thanks!

tsalo:
@chiuhoward I think just having the age in your participants.tsv should be enough, but I don't think we've tested the infant processing outside of an HBCD-specific context. We didn't add `--age-months` because I think it's already deprecated in Nibabies.

When we implement `--output-spaces` users will be able to specify the cohort in the template string, but in the meantime specifying the age in the TSV is probably the only way to get it working.

PR #859, opened 2024-11-11
Change T1wACPC to ACPC and T1wNative to anat

### LINKED ISSUE #853: Rename ACPC-aligned T1w/T2w space to something that is modality-independent
## Summary

@mattcieslak and I have been discussing output naming and he noted that have "space-T1w" and "space-T2w" as the primary anatomical output space depending on the modality of the anatomical data is confusing and makes ingression difficult. We should use a name that is modality independent. Here are a few options we discussed:

- space-ACPC/acpc: This label is used for an iEEG coordinate system that is similar to what QSIPrep has. See https://bids-specification.readthedocs.io/en/stable/appendices/coordinate-systems.html#ieeg-specific-coordinate-systems
- space-anat: Because the space is derived from one or more anatomical images.
- space-individual: This is what BIDS recommends for a subject-specific anatomical space. See https://bids-specification.readthedocs.io/en/stable/appendices/coordinate-systems.html#nonstandard-coordinate-system-identifiers
- space-individualACPC: To clarify that the individual anatomical space is ACPC-aligned.
- space-anatACPC: To clarify that the individual anatomical space is ACPC-aligned.

Let's translate this to a vote. Please respond with the following:

👍  for space-ACPC
👎  for space-acpc
😄 for space-anat
🎉 for space-individual
😕 for space-individualACPC
❤️ for space-anatACPC
🚀 if you feel strongly that we need something else entirely

Vote for as many as you think are fine and we'll take the one with the highest count at the end. We can close the vote on Monday morning so that we have a few days to implement the change in time for the HBCD version release.

tsalo:
Closes #853.


.circleci/DSCSDSI_outputs.txt
```diff
@@ -26,13 +26,13 @@ qsiprep/sub-tester/dwi/sub-tester_acq-HASC55AP_confounds.tsv
 qsiprep/sub-tester/dwi/sub-tester_acq-HASC55AP_desc-ImageQC_dwi.csv
 qsiprep/sub-tester/dwi/sub-tester_acq-HASC55AP_desc-SliceQC_dwi.json
 qsiprep/sub-tester/dwi/sub-tester_acq-HASC55AP_dwiqc.json
```
mattcieslak, line 28: Do you think we should add space- tags to these too? Some of the QC metrics are calculated on the ACPC-resampled data

.circleci/DSDTI_outputs.txt
```diff
@@ -6,15 +6,15 @@ qsiprep/logs/CITATION.md
...
 qsiprep/sub-PNC/anat
-qsiprep/sub-PNC/anat/sub-PNC_desc-brain_mask.nii.gz
-qsiprep/sub-PNC/anat/sub-PNC_desc-preproc_T1w.nii.gz
-qsiprep/sub-PNC/anat/sub-PNC_dseg.nii.gz
+qsiprep/sub-PNC/anat/sub-PNC_space-ACPC_desc-brain_mask.nii.gz
+qsiprep/sub-PNC/anat/sub-PNC_space-ACPC_desc-preproc_T1w.nii.gz
+qsiprep/sub-PNC/anat/sub-PNC_space-ACPC_dseg.nii.gz
 qsiprep/sub-PNC/anat/sub-PNC_from-MNI152NLin2009cAsym_to-T1w_mode-image_xfm.h5
 qsiprep/sub-PNC/anat/sub-PNC_from-orig_to-T1w_mode-image_xfm.txt
 qsiprep/sub-PNC/anat/sub-PNC_from-T1w_to-MNI152NLin2009cAsym_mode-image_xfm.h5
```
mattcieslak, line 14: Should this be `from-ACPC_to-MNI`?

docs/preprocessing.rst
```diff
@@ -231,35 +227,34 @@ is a sampling scheme from a DSI scan:
...
+- ``<source_entities>_space-ACPC_desc-brain_mask.nii.gz`` Brain mask derived using ANTs' ``antsBrainExtraction.sh``.
+- ``<source_entities>_space-ACPC_label-CSF_probseg.nii.gz``
+- ``<source_entities>_space-ACPC_label-GM_probseg.nii.gz``
+- ``<source_entities>_space-ACPC_label-WM_probseg.nii.gz`` tissue-probability maps.
+- ``<source_entities>_space-ACPC_dseg.nii.gz`` Tissue class map derived using FAST.
+- ``<source_entities>_space-ACPC_desc-preproc_T1w.nii.gz`` Bias field corrected T1w file, using ANTS' N4BiasFieldCorrection
+- ``<source_entities>_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz`` Same as ``_brainmask`` above, but in MNI space.
+- ``<source_entities>_space-MNI152NLin2009cAsym_label-CSF_probseg.nii.gz``
+- ``<source_entities>_space-MNI152NLin2009cAsym_label-GM_probseg.nii.gz``
+- ``<source_entities>_space-MNI152NLin2009cAsym_label-WM_probseg.nii.gz`` Probability tissue maps, transformed into MNI space
```
mattcieslak, line 247: this is a great addition

mattcieslak (commented):
Only a few questions

.circleci/DSCSDSI_outputs.txt
tsalo: I think it's reasonable to include it. I'll update accordingly.

.circleci/DSDTI_outputs.txt
tsalo: Ah right, I didn't consistently update the files in `.circleci`. I'm not actually sure which ones are still being used. A bunch of them are duplicates of the tests in `test_cli.py`, right?

mattcieslak: 🤦 I didn't see the directory this is in. We should remove most of this directory at some point, since it doesn't work anymore

qsiprep/tests/data/maternal_brain_project_outputs.txt
```diff
@@ -7,29 +7,29 @@ logs/CITATION.tex
...
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-3dSHORE_cnr.nii.gz
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-brain_mask.nii.gz
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-preproc_dwi.b
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-preproc_dwi.bval
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-preproc_dwi.bvec
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-preproc_dwi.nii.gz
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-preproc_dwi.json
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_desc-preproc_dwi.b_table.txt
-sub-01/ses-01/dwi/sub-01_ses-01_space-T1w_dwiref.nii.gz
+sub-01/ses-01/dwi/sub-01_ses-01_space-ACPC_desc-3dSHORE_cnr.nii.gz
```
tsalo, line 25: Just a note for the future- we might want to replace the `cnr` suffix with `stat-CNR_dwimap`.

qsiprep/workflows/dwi/finalize.py
```diff
@@ -343,6 +343,7 @@ def init_dwi_finalize_wf(
     series_qc = pe.Node(SeriesQC(output_file_name=output_prefix), name="series_qc")
     ds_series_qc = pe.Node(
         DerivativesDataSink(
+            space="ACPC",
             desc="ImageQC",
```
mattcieslak, line 347: I wonder if it makes sense to have desc-ImageQC, when it has the stats for desc-preproc_dwi. wdyt?

mattcieslak (approved):
looks good to go!

qsiprep/workflows/dwi/finalize.py
tsalo: I don't think the `desc` entity needs to match up, but at least in XCP-D I have `desc-linc_qc` instead of `desc-LINCQC_bold`. So what about `desc-image_qc` and `desc-slice_qc`?

mattcieslak: I like that much more. The capitalization of the previous name is also very non-bidsy

tsalo: Changed!

tsalo:
Something went wrong with the filename patterns for `desc-slice_qc`. I'll look into it tomorrow.

tsalo:
I always assumed that the filename patterns in the DerivativesDataSink were selected based on the number of matching entities, but I guess the order of patterns in the list actually matters. Moving the "bad" pattern to the end of the list seems to fix the problem.

tsalo:
It's finally passing. @mattcieslak could you take one more look before we merge?

mattcieslak (approved):
I can add those small docs changes to a different PR

tsalo:
Thanks! I'll merge once CI passes then.

PR #861, opened 2024-11-11
Remove underscore versions of all parameters

tsalo:
Closes none. This is a breaking change because parameter names have effectively been changed.

- Remove underscore versions of all parameters (e.g., `--anat_only` instead of `--anat-only`).
- Add dash versions where they were missing.


mattcieslak (approved):
didn't fmriprep get rid of the underscore versions? I'm ok with keeping them here

tsalo:
Oh you're right! I might have corrected in the wrong direction. Do you have any issue with me reversing this and dropping the underscore versions?

mattcieslak:
I trust the fmriprep folks on this decision

tsalo:
Okay cool, I'll reverse my changes.

mattcieslak (approved):

PR #862, opened 2024-11-11
Make grouping account for acquisition entity

### LINKED ISSUE #860: Limit distortion groups to a single `acq` entity value
## Summary

This will make the distortion group definition more restrictive. @mattcieslak and I discussed it, and we don't think this will affect many datasets, but will make it so that methods-focused datasets (i.e., ones with different DWI sequences in a given session) can be run through QSIPrep once, without having to create BIDS filter files to select a given `acq` value and re-run multiple times.

tsalo:
Closes #860.

- Rename `get_session_groups` to `get_entity_groups`.
- Generalize `get_entity_groups` to group scans within both session _and_ acquisition. The addition of acquisition is new.

### DIFF
--- a/qsiprep/utils/grouping.py
+++ b/qsiprep/utils/grouping.py
@@ -114,8 +114,9 @@ def get_entity_groups(layout, subject_data, combine_all_dwis):
             ]
             for acq in acquisitions:
                 group_files = [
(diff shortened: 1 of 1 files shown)


mattcieslak (approved):
Brilliant!

PR #875, opened 2024-11-14
Move common metadata to top of JSON

tsalo:
Closes none.

- Compare metadata dictionaries across runs in a merge group and compile the metadata fields that are consistent across runs. This then goes in the merged JSON file. QSIRecon can reference the metadata fields at the top level of the JSON.
    - In cases where the constituent runs don't have the same value for a given field, that field will be represented in the "SourceMetadata" field, but not in the top level of the JSON.
- Rename "scan_grouping" field to "ScanGrouping".  **THIS IS A BREAKING CHANGE**
- Rename "source_metadata" field to "SourceMetadata". **THIS IS A BREAKING CHANGE**
- Add "Sources" field to the JSON. I want to change all of the paths to BIDS-URIs in the future, but it's not a high priority item.

### DIFF
--- a/qsiprep/utils/bids.py
+++ b/qsiprep/utils/bids.py
@@ -435,7 +435,8 @@ def scan_groups_to_sidecar(scan_groups):
     """Create a sidecar that reflects how the preprocessed image was created."""
 
     # Add the information about how the images were grouped and which fieldmaps were used
-    derivatives_metadata = {"scan_grouping": scan_groups}
+    derivatives_metadata = {"ScanGrouping": scan_groups}
+    derivatives_metadata["Sources"] = scan_groups["dwi_files"]
 
     # Get metadata from the individual scans that were combined to make this preprocessed image
     concatenated_dwi_files = scan_groups.get("dwi_series")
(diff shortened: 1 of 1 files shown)


qsiprep/utils/bids.py
```diff
@@ -435,18 +435,41 @@ def scan_groups_to_sidecar(scan_groups):
...
+                            keys_to_remove.append(key)
+                    elif val1 != val2:
+                        keys_to_remove.append(key)
+
+            for key in keys_to_remove:
+                common_metadata.pop(key)
+
+    derivatives_metadata["SourceMetadata"] = scan_metadata
+    derivatives_metadata = {**common_metadata, **derivatives_metadata}
+
```
mattcieslak, line 472:
```suggestion
derivatives_metadata["Sources"] = sorted(scan_metadata.keys())
```

qsiprep/utils/bids.py
```diff
@@ -435,18 +435,41 @@ def scan_groups_to_sidecar(scan_groups):
     """Create a sidecar that reflects how the preprocessed image was created."""
 
     # Add the information about how the images were grouped and which fieldmaps were used
-    derivatives_metadata = {"scan_grouping": scan_groups}
+    derivatives_metadata = {"ScanGrouping": scan_groups}
+    derivatives_metadata["Sources"] = scan_groups["dwi_series"]
```
mattcieslak, line 439:
```suggestion

```

mattcieslak (approved):
One quickfix but otherwise this is perfect

qsiprep/utils/bids.py
tsalo: The keys are filenames rather than full paths, right?

tsalo: I think they're just the filenames, not relative paths.

PR #880, opened 2024-11-18
Hardcode extensions for GIF figures

tsalo:
This is just until the next nireports release is made, since GIFs aren't supported yet and that's breaking the HTML report generation.

- Hardcode extensions for figures that are actually GIFs as PNGs so they get skipped until nireports can support GIFs.


mattcieslak:
Will this still place the gifs in the figures directory? If so then I like this solution

tsalo:
The figures will still be created, but nireports won't be able to find them because it'll be looking for PNGs (which don't exist) instead. Then when the new release goes out, we can just drop the ".png" extension part and it should find the GIFs.

Issue #681, opened 2024-01-26
Replace `--anatomical-template` and `--output-resolution` with `--output-spaces`


tsalo:
This would bring QSIPrep more in line with other preps, and would maybe make it easier to support a `res-native` option for cases where resolution varies across acquisitions, sessions, etc.

@mattcieslak pointed out that most recon methods require isotropic voxels, so we might need something like `res-nativemax` and `res-nativemin` to use the largest or smallest voxel dimension, respectively.

For example, if voxels are 1.7 x 1.7 x 1.718, `res-nativemax` would produce outputs with 1.718 x 1.718 x 1.718 voxels and `res-nativemin` would produce outputs with 1.7 x 1.7 x 1.7 voxels.

tsalo:
Satra proposed something like `res-isoXmm` for arbitrary isotropic resolutions in https://github.com/nipreps/fmriprep/issues/2424#issuecomment-876064758.

tsalo:
We should _require_ ACPC as an output space, and other output spaces won't use the resolution component.

So you could do something like:

```
--output-spaces ACPC:res-2:res-nativemin:res-nativemax MNI152NLin2009cAsym MNI152NLin6Asym
```

That would output preprocessed DWI in ACPC with three different resolutions, and would also generate transforms to MNI152NLin6Asym and MNI152NLin2009cAsym.

tsalo:
I think this proposal can take over for #620. We could allow other output spaces than ACPC and raise a warning/error if someone wants something that doesn't make sense.

tsalo:
It would simplify the interface _and_ make QSIRecon's data ingression easier. We can always allow QSIRecon to ingress derivatives from previous QSIRecon runs (see https://github.com/PennLINC/qsirecon/issues/111), so the transforms wouldn't need to be recalculated.

I just worry about compatibility with other post-processing workflows. Minimal preprocessing should preferably produce self-contained outputs. It would be good to check with some of the other dMRI folks to make sure QSIPrep derivatives would still be useful without thoe transforms.

mattcieslak:
As of now, those transforms are _only_ used for the connectivity matrices. They used to be used for PyAFQ as well, but a recent PyAFQ update made the format incompatible with PyAFQ, so it just recalculates its own transform.

mattcieslak:
correction: connectivity matrices and getting the ROIs for the ODF plots in the QSIRecon reports

WilliamFCB:
Hi

[...]

PR #889, opened 2024-11-22
Remove unused transform connections

tsalo:
Closes none. I'm attempting to remove any unused connections that might interfere with #681.

- Remove template-related connections from DWI processing workflows.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -160,12 +160,14 @@ ignore = [
   "DTZ005",  # TODO: Fix these
   "B904",  # TODO: Fix these
   "A001",  # TODO: Fix these
+  "A004",  # TODO: Fix these
   "B006",  # TODO: Fix these
   "S607",  # TODO: Fix these
   "S108",  # TODO: Fix these
   "S602",  # TODO: Fix these
   "E402",  # TODO: Fix these
   "UP028",  # TODO: Fix these
+  "UP031",  # TODO: Fix these
   "BLE001",  # TODO: Fix these
 ]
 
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -160,12 +160,14 @@ ignore = [
...
   "A001",  # TODO: Fix these
+  "A004",  # TODO: Fix these
   "B006",  # TODO: Fix these
   "S607",  # TODO: Fix these
   "S108",  # TODO: Fix these
   "S602",  # TODO: Fix these
   "E402",  # TODO: Fix these
   "UP028",  # TODO: Fix these
+  "UP031",  # TODO: Fix these
   "BLE001",  # TODO: Fix these
```
tsalo, line 171: I don't know why new style issues are suddenly cropping up. It looks like ruff was just missing them before. My local version isn't catching them either.

qsiprep/workflows/fieldmap/base.py
```diff
@@ -264,16 +260,17 @@ def init_sdc_wf(fieldmap_info, dwi_meta):
...
             bold_pe=dwi_meta.get('PhaseEncodingDirection', None),
         )
 
         workflow.connect([
             (inputnode, syn_sdc_wf, [
                 ('t1_brain', 'inputnode.t1_brain'),
+                # Only place a template transform is used on DWI data
                 ('t1_2_mni_reverse_transform', 'inputnode.t1_2_mni_reverse_transform'),
                 ('b0_ref', 'inputnode.bold_ref'),
-                ('template', 'inputnode.template'),
```
tsalo, line 276: I just want to note this. When we switch to `--output-spaces`, we should (1) raise an error if users try to use syn with infant data and (2) only pass along the 2009c template transform here.

tsalo:
This makes me think that we should track through the workflow connections across the package and drop any that are unused. @mattcieslak WDYT?

mattcieslak:
I think we _will_ actually need most of these once we add the synb0-disco method

tsalo:
I thought SynB0's requirements were basically the same as syn's. Is that not the case?

PR #892, opened 2024-11-26
Calculate patch radius from number of volumes in DWI run

### LINKED ISSUE #845: Denoising patch radius clarifications
## Summary

In `v. 0.22.0` and forward, the behavior of `--dwi-denoise-window` changed. Previously, not setting this parameter (thus `auto`) set `dwidenoise` to run with an extent window of 5x5x5 and `patch2self` to run with a radius that depended on the number of volumes but for adequately sized datasets (number of non-b0 volumes > 50) the default patch radius = 0x0x0. In `v 0.22.0` and forward, `dwidenoise` approach has not changed but `patch2self` now gives an error because the variable `dwi_denoise_window` is no longer set prior to the call (it is only created when using `dwidenoise` and `auto` for the window). See #827.

Developer recommendations for `dwidenoise` and `patch2self` suggest the following about the extent window/patch radius:

1. `dwidenoise` by default will select a patch size that minimally exceeds the number of diffusion weighted image: "By default, the command will select the smallest isotropic patch size that exceeds the number of DW images in the input data, e.g., 5x5x5 for data with <= 125 DWI volumes, 7x7x7 for data with <= 343 DWI volumes, etc." (https://mrtrix.readthedocs.io/en/dev/reference/commands/dwidenoise.html)
2. From @ShreyasFadnavis in #827, the recommended patch radius for `patch2self` is 0x0x0 for human data, though infant and NHP (and certainly small animal) data may better use 1x1x1 or larger.

In light of these two recommendations, the question is: How much control is desired over the patch radius? There are two (probably more) relatively straightforward alternatives:
1. **Allow users to have always and complete control over the patch radius.** When not specified, "auto" is fixed to correctly propagate to both `dwidenoise` and `patch2self` to allow for automatic selection (`dwidenoise`) or a default 0x0x0 (`patch2self`). When specified, it is passed to the relevant denoising function as an isotropic patch size.
2. **Control over patch2self only.** Because `dwidenoise` will choose a sane default value, altering the patch radius may have unintended consequences if too large or too small relative to the number of volumes. However, there may be useful experimentation with the patch radius for `patch2self` (or in cases of infant/NHP data) where being able to set a non-default value is reasonable and possibly necessary.
3. A third option would be to allow no control, always allowing `dwidenoise` to auto-select the extent window and using a 0x0x0 patch radius for human data (and maybe a 1x1x1 for infant data).

Looking for thoughts before working on a solution to the current error where `patch2self` is not working at all.

@mattcieslak feel free to update or clarify as needed.

tsalo:
Closes #845.

- When using `--denoise-method dwidenoise` and `--dwi-denoise-window auto`, replace default value of 5 with what we _think_ dwidenoise's default value will be based on the number of volumes in the DWI run.
- Ignore `--dwi-denoise-window` when using `patch2self`.
    - Raise an error at the parsing step if a user selects a non-default value for `--dwi-denoise-window` _and_ either `--denoise-method none` or `--denoise-method patch2self`.
- Check that `--dwi-denoise-window` is a positive, odd integer when it is supplied.
- Update the boilerplate to include the new dwi-denoise-window value.

### DIFF
--- a/qsiprep/cli/parser.py
+++ b/qsiprep/cli/parser.py
@@ -168,6 +168,7 @@ def _build_parser(**kwargs):
     PathExists = partial(_path_exists, parser=parser)
     IsFile = partial(_is_file, parser=parser)
     PositiveInt = partial(_min_one, parser=parser)
+    IntOrAuto = partial(_int_or_auto, parser=parser)
     BIDSFilter = partial(_bids_filter, parser=parser)
 
     # Arguments as specified by BIDS-Apps
@@ -363,7 +364,7 @@ def _build_parser(**kwargs):
     g_conf.add_argument(
         '--dwi-denoise-window',
         action='store',
-        type=_int_or_auto,
+        type=IntOrAuto,
         default='auto',
         help=(
             'Window size in voxels for image-based denoising: integer or "auto". '
--- a/qsiprep/cli/parser.py
+++ b/qsiprep/cli/parser.py
@@ -99,6 +99,20 @@ def _build_parser(**kwargs):
             raise parser.error("Argument can't be less than one.")
         return value
 
+    def _int_or_auto(value, parser):
+        """Ensure an argument is an integer or 'auto'."""
+        if value == 'auto':
+            return value
+        try:
+            value = int(value)
+        except ValueError as exc:
+            raise parser.error('Argument must be an integer or "auto".') from exc
+
+        if value < 1:
+            raise parser.error('Argument must be greater than zero.')
+
+        return value
+
     def _to_gb(value):
         scale = {'G': 1, 'T': 10**3, 'M': 1e-3, 'K': 1e-6, 'B': 1e-9}
         digits = ''.join([c for c in value if c.isdigit()])
@@ -349,10 +363,16 @@ def _build_parser(**kwargs):
     g_conf.add_argument(
         '--dwi-denoise-window',
         action='store',
+        type=_int_or_auto,
         default='auto',
-        help='window size in voxels for image-based denoising, integer or "auto".'
-        'If "auto", 5 will be used for dwidenoise and auto-configured for '
-        'patch2self based on the number of b>0 images.',
+        help=(
+            'Window size in voxels for image-based denoising: integer or "auto". '
+            'If using the "dwidenoise" denoising method, '
+            'the "auto" option will calculate a window size '
+            'based on the number of volumes according to the method described by the '
+            'dwidenoise documentation. '
+            'If using the "patch2self" denoising method, this argument will not be used.'
+        ),
     )
     g_conf.add_argument(
         '--denoise-method',
@@ -360,7 +380,7 @@ def _build_parser(**kwargs):
         choices=['dwidenoise', 'patch2self', 'none'],
         default='dwidenoise',
         help='Image-based denoising method. Either "dwidenoise" (MRtrix), '
-        '"patch2self" (DIPY) or none. (default: dwidenoise)',
+        '"patch2self" (DIPY) or "none". (default: dwidenoise)',
     )
     g_conf.add_argument(
         '--unringing-method',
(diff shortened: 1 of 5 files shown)


tsalo:
I still need to log the patch radius in the boilerplate.

qsiprep/workflows/dwi/merge.py
```diff
@@ -149,13 +149,13 @@ def init_merge_and_denoise_wf(
 
     # Get a data frame of the raw_dwi_files and their imaging parameters:
     dwi_df = get_acq_parameters_df(raw_dwi_files, layout=layout)
-    for dwi_num, row in dwi_df.iterrows():
-        dwi_num += 1  # start at 1
+    for i_dwi, row in dwi_df.iterrows():
+        dwi_num = i_dwi + 1  # start at 1
```
tsalo, line 153: I just changed this because modifying a loop variable in the loop is considered bad practice. I don't actually think it would have any effect here.

qsiprep/cli/parser.py
```diff
@@ -349,18 +364,24 @@ def _bids_filter(value, parser):
     g_conf.add_argument(
         '--dwi-denoise-window',
         action='store',
+        type=IntOrAuto,
         default='auto',
-        help='window size in voxels for image-based denoising, integer or "auto".'
-        'If "auto", 5 will be used for dwidenoise and auto-configured for '
-        'patch2self based on the number of b>0 images.',
+        help=(
+            'Window size in voxels for image-based denoising: integer or "auto". '
```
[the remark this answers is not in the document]
tsalo: I added a check.

qsiprep/workflows/dwi/merge.py
```diff
@@ -438,7 +443,30 @@ def get_buffernode():
...
 
+        dwi_denoise_window = config.workflow.dwi_denoise_window
+        auto_str = ''
+        if denoise_method == 'dwidenoise' and config.workflow.dwi_denoise_window == 'auto':
+            # Configure the denoising window
+            import numpy as np
+
+            dwi_denoise_window = closest_odd(int(np.cbrt(n_volumes)))
+            config.loggers.workflow.info(
+                f'Automatically using {dwi_denoise_window}, {dwi_denoise_window}, '
```
mattcieslak, line 454: What does this boilerplate look like if it's independently denoising a bunch of dwis?

qsiprep/workflows/dwi/merge.py
```diff
@@ -734,3 +749,10 @@ def get_merged_parameter(parameter_df, parameter_name, selection_mode='all'):
         return col.mode()[0]
 
     raise Exception("selection_mode must be 'all' or 'mode'")
+
+
+def closest_odd(x):
```
mattcieslak, line 754: I believe patch2self supports a patch radius of 0

qsiprep/cli/parser.py
```diff
@@ -99,6 +99,23 @@ def _min_one(value, parser):
...
+        try:
+            value = int(value)
+        except ValueError as exc:
+            raise parser.error('Argument must be an integer or "auto".') from exc
+
+        if value < 1:
+            raise parser.error('Argument must be greater than zero.')
+
+        if value % 2 == 0:
+            raise parser.error('Argument must be an odd integer.')
```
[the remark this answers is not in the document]
tsalo: We don't use the value for patch2self at all though, so the only time `--dwi-denoise-window` has an effect is if you use dwidenoise.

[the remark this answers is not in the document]
mattcieslak: ah, so this doesn't even matter here

qsiprep/workflows/dwi/merge.py
tsalo: I think it'll have separate copies of very similar paragraphs. Users should modify their boilerplates for publication in that case though.

qsiprep/cli/parser.py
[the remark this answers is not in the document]
tsalo: Yeah, the parser is currently going to raise an error if a user provides a non-auto value and _isn't_ using dwidenoise, since it won't have an effect.

qsiprep/workflows/dwi/merge.py
mattcieslak: My bad, this is irrelevant

mattcieslak (approved):

PR #895, opened 2024-12-03
fix mount instructions for micromamba

mattcieslak:
An old branch, checking to see if this is still needed


mattcieslak:
Not relevant anymore, path changed

PR #901, opened 2024-12-09
Fix session processing for datasets with mixed modalities

tsalo:
Closes none, but addresses two bugs identified by @araikes.

- Directly merge "session_id" (from the processing list) into the BIDS queries (which include the BIDS filters, which in turn may include "session") within `collect_data`.
    - This should fix a bug that comes from using `--subject-anatomical-reference sessionwise` in combination with a `--bids-filter-file` that includes `session` fields.
- Limit session-wise processing to sessions with DWI data.
    - This should fix a bug that comes from using `--subject-anatomical-reference sessionwise` on a dataset where some sessions have anatomical data without a corresponding DWI scan.

### DIFF
--- a/qsiprep/cli/parser.py
+++ b/qsiprep/cli/parser.py
@@ -800,10 +800,11 @@ def parse_args(args=None, namespace=None):
 
     # Examine the available sessions for each participant
     for subject_id in participant_label:
+        # Find sessions with DWI data
         sessions = config.execution.layout.get_sessions(
             subject=subject_id,
             session=session_filters or Query.OPTIONAL,
-            suffix=['T1w', 'T2w', 'dwi'],
+            suffix=['dwi'],
         )
 
         # If there are no sessions, there is only one option:
@@ -811,7 +812,7 @@ def parse_args(args=None, namespace=None):
             if config.workflow.subject_anatomical_reference == 'sessionwise':
                 config.loggers.workflow.warning(
                     f'Subject {subject_id} had no sessions, '
-                    "but --subject-anatomical-reference was set to 'sessionwise'. "
+                    'but --subject-anatomical-reference was set to "sessionwise". '
                     'Outputs will NOT appear in a session directory for '
                     f'{subject_id}.',
                 )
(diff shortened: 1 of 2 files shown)


qsiprep/cli/parser.py
```diff
@@ -823,6 +824,12 @@ def parse_args(args=None, namespace=None):
             for session in sessions:
                 processing_groups.append([subject_id, [session]])
         else:
+            # We can now use sessions that have anatomical data, but no DWI
```
tsalo, line 827: This will process "unbiased" or "first" options using all of the anatomical data, I think.

qsiprep/utils/bids.py
```diff
@@ -206,14 +206,14 @@ def collect_data(bids_dir, participant_label, session_id=None, filters=None, bid
...
     for acq, entities in bids_filters.items():
+        queries[acq]['session'] = session_id or Query.OPTIONAL
         queries[acq].update(entities)
 
     subj_data = {
         dtype: sorted(
             layout.get(
                 return_type='file',
                 subject=participant_label,
-                session=session_id or Query.OPTIONAL,
```
tsalo, line 216: This was conflicting with `**query` when `query` had a "session" key.

mattcieslak: will this overwrite the bids filter file value with the command line arg? I'm ok with this behavior but we should warn that the bids filter file value is being overwritten

qsiprep/utils/bids.py
```diff
@@ -206,14 +206,14 @@ def collect_data(bids_dir, participant_label, session_id=None, filters=None, bid
     }
     bids_filters = filters or {}
     for acq, entities in bids_filters.items():
+        queries[acq]['session'] = session_id or Query.OPTIONAL
```
mattcieslak, line 209:
```suggestion
        if 'session' in queries[acq]:
            config.loggers.workflow.warning('BIDS filter file value for session may conflict with values specified on the command line')
        queries[acq]['session'] = session_id or Query.OPTIONAL
```

qsiprep/cli/parser.py
mattcieslak: This looks right

qsiprep/utils/bids.py
tsalo: No, it should override the session from the processing_list with the one from the filter instead.

mattcieslak (approved):

mattcieslak (approved):

PR #902, opened 2024-12-09
Use DIPY version of patch2self

### LINKED ISSUE #893: Use official DIPY version of patch2self
We added patch2self before it was officially released in DIPY, so there has been a copy of it in qsiprep for some time. 

@ShreyasFadnavis can you think of any reason why we'd need to keep a copy of patch2self in qsiprep instead of using the one in DIPY?

tsalo:
Closes #893.

- Remove internal implementation of `patch2self` and use version imported from dipy.
- Fix Patch2Self interface to retain noise image and nmse text file in outputs.


qsiprep/interfaces/denoise.py
```diff
@@ -62,6 +67,7 @@ def _calculate_nmse(self, original_nii, corrected_nii):
         pd.DataFrame(
             {title + '_pre': pres, title + '_post': posts, title + '_change': differences}
         ).to_csv(output_file, index=False)
+        self._nmse_text = output_file
```
mattcieslak, line 70: This is really clever!

mattcieslak (approved):
Great solution! I think we might be running some redundant testing now, but that's better than the alternative

PR #903, opened 2024-12-10
ENH: Update eddy_params.json

cookpa:

Change the default eddy flm from linear to quadratic. This matches the eddy default.

The slm remains linear, as recommended for lower-resolution data.

I don't think the defaults are mentioned in the docs.


mattcieslak (approved):
I agree completely. Thanks for making the PR

PR #909, opened 2025-01-09
Fix use of `session_id` in `collect_data`

### LINKED ISSUE #908: Session-wise processing is not actually session-wise
## Summary

In a multi-session dataset, running QSIPrep with `--subject-anatomical-reference sessionwise` does not correctly filter the dataset, resulting in duplicate processing and non-session-specifc outputs 

## Additional details

- QSIPrep version: 1.0.0rc2.dev12+g716aac4.d20241211
- Apptainer version: 1.3.2.-1.el7

Session-wise processing of a multi-session dataset.

I expected `ses-01` to be processed with only the T1w, DWI, and reverse phase EPI from `ses-01` and like for `ses-02` to only process files from that session.

### What actually happened?
Within the work directory, two top level folders were created, as expected `sub-001_ses-01_wf` and `sub-001_ses-02_wf`.

Several things happened that were not as expected. `sub-001_ses-01_wf` will be used as the exemplar but it is identically applicable to the other top level workflow folder.
1. Instead of only processing the `ses-01` T1w image, the unbiased reference work flow was run, such that what should have been the session-specific preprocessed image was instead an unbiased average image of multiple sessions.
2. The DWI files from both `ses-01` and `ses-02` were processed in parallel without merging, whereas this workflow should only have processed `ses-01` while `ses-02` was processed in `sub-001_ses-02_wf`. This has the effect of duplicating all of the processing steps.

The output DWI files appeared to be correct (correct file names, no duplication or sessions in unexpected locations), but it wasn't clear which files were ultimately being written from the workflows to the output.

Including a `bids-filter-file` to limit session selection along with specifying `--session-id` appears to correct this. Only specifying `--session-id` replicates this behavior but only generates the specified top-level workflow. Thus if `--session-id 01` is passed, then all of the issues are still present but `sub-001_ses-02_wf` was not created. 

## Reproducing the bug

Initial call:
```
apptainer run --nv \
--containall \
-B /xdisk/adamraikes/PD_allo/nifti:/nifti:ro \
-B /xdisk/adamraikes/PD_allo/derivatives/qsiprep_1.0.0rc2_redo:/xdisk/adamraikes/PD_allo/derivatives/qsiprep_1.0.0rc2_redo \
-B /xdisk/adamraikes/PD_allo/extra:/extra \
-B /groups/adamraikes/templateflow:/opt/templateflow \
-B /groups/adamraikes/license.txt:/license.txt \
-B /tmp:/tmp \
/groups/adamraikes/singularity_images/qsiprep_1.0.0rc2.sif \
/nifti /xdisk/adamraikes/PD_allo/derivatives/qsiprep_1.0.0rc2_redo participant \
--participant-label 001 \
--nthreads 24 --omp-nthreads 8 \
--output-resolution 1 \
--subject-anatomical-reference sessionwise \
--denoise-method patch2self \
--unringing-method rpg \
--eddy-config /extra/eddy_params_v2.json \
--fs-license-file /license.txt \
-w /xdisk/adamraikes/PD_allo/derivatives/qsiprep_1.0.0rc2_redo/scratch \
--skip-bids-validation -vv
```

tsalo:
Closes #908. `collect_data` wasn't using `session_id` except when there's a BIDS filter field.

- Fix `collect_data`'s handling of `session_id`.
- Add tests using simulated datasets.


qsiprep/utils/bids.py
```diff
@@ -226,6 +230,11 @@ def collect_data(bids_dir, participant_label, session_id=None, filters=None, bid
         for dtype, query in queries.items()
     }
 
+    config.loggers.workflow.log(
+        25,
+        f'Collected data:\n{yaml.dump(subj_data, default_flow_style=False, indent=4)}',
```
mattcieslak, line 235: nice!

mattcieslak (approved):
THANK YOU

PR #914, opened 2025-01-13
Set output directory for derivative datasinks

tsalo:
Closes none.

- Assign the `base_directory` for any datasinks to the output directory.
- Change `desc-image_qc.csv` to `desc-image_qc.tsv`. This is a breaking change.
- Write out `topupcsv.csv` (now called `desc-pepolar_qc.tsv`) to output directory.


tsalo:
I just need to fix the datasink for the topup qc file.

qsiprep/tests/data/dsdti_topup_outputs.txt
mattcieslak, line 1: is the pepolar_qc supposed to appear in this one?

tsalo: It looks like dsdti_topup is defined in the pytest markers but there's no corresponding test in the `tests/` folder.

tsalo: Also, is it weird that dsdti_nofmap is creating this file?

mattcieslak: I think in the previous setup, DSDTI_nofmap downloaded data that had fieldmaps, then deleted those fieldmaps in a script. So this new nofmap is actually running with a pepolar fieldmap

tsalo: We should probably run both versions of the test and add a BIDS filter file to ignore the field maps in the nofmap test.

tsalo: I'd rather push that for a separate PR though.

tsalo: Just opened #916 for this.

PR #918, opened 2025-02-19
Group DWI scans by MultipartID when available

### LINKED ISSUE #897: Add option to revert to old behaviour of grouping scans across different acquisitions?
## Summary

Hi, I have a dataset where some of the participants have multi-shell data split across multiple scans (one scan per b-value), and we used the `acq` entity to label them. For example the raw BIDS data would look something like this:
1. `sub-001_ses-1_acq-B0_dir-AP_run-01_dwi.nii.gz`
2. `sub-001_ses-1_acq-B0_dir-AP_run-02_dwi.nii.gz`
3. `sub-001_ses-1_acq-B0_dir-AP_run-03_dwi.nii.gz`
4. `sub-001_ses-1_acq-B0_dir-AP_run-04_dwi.nii.gz`
5. `sub-001_ses-1_acq-B0_dir-PA_run-01_dwi.nii.gz`
6. `sub-001_ses-1_acq-B1000_dir-PA_run-01_dwi.nii.gz`
7. `sub-001_ses-1_acq-B2000_dir-PA_run-01_dwi.nii.gz`
8. `sub-001_ses-1_acq-B700_dir-PA_run-01_dwi.nii.gz`

So there are four B0 scans in the AP direction, then 4 in the PA direction with b-values 0, 1000, 2000 and 700. All of them have accompanying `.bval`/`.bvec` files too.

PR #862 changed QSIPrep's behaviour to account for the `acq` tag during the distortion group definition. Unfortunately, this results in undesirable behaviour in 1.0.0rc1 in my case, where the scans are not being merged as expected. With 0.23.0 I was getting a single `desc-preproc_dwi.nii.gz` file, while with 1.0.0rc1 I am getting a `desc-preproc_dwi.nii.gz` file for each of the B1000, B2000, and B700 scans, and I think there was was an error for the B0 scans because only the JSON was produced (might be a related issue).

I understand the reasoning for grouping by `acq`, but would it be possible to add a flag that allows the user to revert to the old behaviour of ignoring the `acq` entity when defining the scan groups? 

Happy to create a PR for this if you could point me to the parts of the code to change.

tsalo:
Closes #897 and reverts #862.

- Modify `get_entity_groups` to use the `MultipartID` metadata field when available.
    - ~~If `MultipartID` is present for some files, but not others, then the function will treat all of the files without the field as belonging to the same group. Now that I think about it, we probably want them to be treated as belonging to separate groups, right? I can make that change depending on what @mattcieslak thinks.~~
    - If `MultipartID` is present for some files, but not others, then the function will treat each file without the field as belonging to its own group.
- Stop grouping by acquisition entity by default. If users want to apply more sophisticated grouping logic, they can use `MultipartID`.
- Test `get_entity_groups`'s  behavior on simulated datasets with and without `MultipartID`.

### DIFF
--- a/qsiprep/tests/test_utils_grouping.py
+++ b/qsiprep/tests/test_utils_grouping.py
@@ -76,18 +76,18 @@ def test_get_entity_groups_with_multipartid(tmpdir):
     entity_groups = grouping.get_entity_groups(layout, subject_data, combine_all_dwis=True)
     expected = [
         [
-            'sub-01_ses-01_dir-AP_acq-99dir_run-01_dwi.nii.gz',
-            'sub-01_ses-01_dir-AP_acq-98dir_run-02_dwi.nii.gz',
+            'sub-01_dir-AP_acq-99dir_run-01_dwi.nii.gz',
+            'sub-01_dir-AP_acq-98dir_run-02_dwi.nii.gz',
         ],
-        ['sub-01_ses-01_dir-AP_acq-99dir_run-03_dwi.nii.gz'],
+        ['sub-01_dir-AP_acq-99dir_run-03_dwi.nii.gz'],
     ]
     check_expected(entity_groups, expected)
 
     entity_groups = grouping.get_entity_groups(layout, subject_data, combine_all_dwis=False)
     expected = [
-        ['sub-01_ses-01_dir-AP_acq-99dir_run-01_dwi.nii.gz'],
-        ['sub-01_ses-01_dir-AP_acq-98dir_run-02_dwi.nii.gz'],
-        ['sub-01_ses-01_dir-AP_acq-99dir_run-03_dwi.nii.gz'],
+        ['sub-01_dir-AP_acq-99dir_run-01_dwi.nii.gz'],
+        ['sub-01_dir-AP_acq-98dir_run-02_dwi.nii.gz'],
+        ['sub-01_dir-AP_acq-99dir_run-03_dwi.nii.gz'],
     ]
     check_expected(entity_groups, expected)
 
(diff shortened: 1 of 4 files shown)


qsiprep/tests/test_utils_grouping.py
mattcieslak, line 1: This is awesome. I'm going to add a bunch more of these in a separate PR

qsiprep/utils/grouping.py
```diff
@@ -90,13 +96,28 @@ def get_entity_groups(layout, subject_data, combine_all_dwis):
...
+    grouping_method = 'entities'  # default is to group by entities
+    if any('MultipartID' in metadata for metadata in all_metadata):
+        grouping_method = 'metadata'
+
+    dwi_entities = {}
+    grouping_metadata = {}
+    for i_file, f in enumerate(all_dwis):
+        if grouping_method == 'metadata':
+            # One MultipartID in any DWI metadata file means we use MultipartID to group
+            # If a DWI has no MultipartID, it will be placed in a group by itself
```
mattcieslak, line 117: Do we want to do this or do we want to group all those without multipart IDs? I think your approach is the right one here - if they use multipart IDs at all, it means that the curator has thought about what should be grouped together and has not specified a group for dwis with no multipart ID.  On the other hand, this is a different behavior than the default for empty multipart IDs. I think this is fine though

mattcieslak (approved):
Much cleaner and straightforward approach to grouping. Thank you!!

qsiprep/utils/grouping.py
tsalo: I think it makes more sense to assume that only specifying MultipartIDs for some files means that the ones without that field should be considered independently, but it _is_ a change from the current behavior. The good thing is that the behavior will stay the same as long as no files have MultipartID.

I see it as somewhat analogous to IntendedFor. If you have IntendedFor pointing from the field maps to some of the DWI files, but other DWIs don't have anything intended for them, you should assume that the field maps shouldn't be applied to them.

mattcieslak: Great point. I agree

Issue #920, opened 2025-02-20
Diffusion run missing from final preprocessed output when using separate epi fieldmaps


valeriejill:
## Summary
When running qsiprep on BIDS data with 2 diffusion runs (both dwi runs in the same PE direction) and 2 b=0 epi fieldmaps (both fieldmap runs in the opposite direction to the dwi data), both diffusion runs appear to get processed, but only run-2 is actually in the preprocessed output file. The 2 diffusion runs were assigned separate b=0 epi fieldmaps.

## Additional details
- QSIPrep version: 1.0.0 (https://hub.docker.com/layers/pennlinc/qsiprep/1.0.0/images/sha256-ebbfe74b9d35df9fbe9873d383a5f11171b0411ae9a9098d330290884c7c853b) 
- Singularity version: singularity-ce version 4.0.1-jammy

### What were you trying to do?

I have BIDS data with the following structure:

```
├── anat
│   ├── sub-12008_ses-1_T1w.json
│   ├── sub-12008_ses-1_T1w.nii.gz
│   ├── sub-12008_ses-1_T2w.json
│   └── sub-12008_ses-1_T2w.nii.gz
├── dwi
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-1_dwi.bval
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-1_dwi.bvec
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-1_dwi.json
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-1_dwi.nii.gz
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-2_dwi.bval
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-2_dwi.bvec
│   ├── sub-12008_ses-1_acq-habit_dir-PA_run-2_dwi.json
│   └── sub-12008_ses-1_acq-habit_dir-PA_run-2_dwi.nii.gz
├── fmap
│   ├── sub-12008_ses-1_acq-b0_dir-AP_run-1_epi.bval
│   ├── sub-12008_ses-1_acq-b0_dir-AP_run-1_epi.bvec
│   ├── sub-12008_ses-1_acq-b0_dir-AP_run-1_epi.json
│   ├── sub-12008_ses-1_acq-b0_dir-AP_run-1_epi.nii.gz
│   ├── sub-12008_ses-1_acq-b0_dir-AP_run-2_epi.bval
[...]

PR #922, opened 2025-02-20
Log steps of concatenation grouping

tsalo:
Closes none.

- Fix up concatenation-related docstrings.
- Log concatenation-related groupings as they're compiled.

### DIFF
--- a/qsiprep/reports/core.py
+++ b/qsiprep/reports/core.py
@@ -24,7 +24,7 @@ from pathlib import Path
 
 from nireports.assembler.report import Report
 
-from .. import data
+from qsiprep import data
 
 
 def run_reports(
--- a/qsiprep/utils/grouping.py
+++ b/qsiprep/utils/grouping.py
@@ -1082,7 +1082,7 @@ def group_for_eddy(all_dwi_fmap_groups):
        'fieldmap_info': {'magnitude1': '.../magnitude1/sub-1/fmap/sub-1_magnitude1.nii.gz',
                          'phasediff': '.../phasediff/sub-1/fmap/sub-1_phasediff.nii.gz',
                          'suffix': 'phasediff'}}],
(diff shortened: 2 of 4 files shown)


qsiprep/utils/grouping.py
```diff
@@ -991,6 +1010,26 @@ def group_for_eddy(all_dwi_fmap_groups):
                '.../mixed_fmaps/sub-1/fmap/sub-1_dir-PA_run-1_epi.nii.gz']},
       'concatenated_bids_name': 'sub-1'}]
 
+    Val's scenario
```
tsalo, line 1013: I think this might be what's happening with @valeriejill's problem (#920).

qsiprep/reports/core.py
```diff
@@ -24,7 +24,7 @@
 
 from nireports.assembler.report import Report
 
-from .. import data
+from qsiprep import data
```
mattcieslak, line 27: does ruff prefer named imports?

qsiprep/utils/grouping.py
mattcieslak: Should we keep this behavior if there are no additional uses of b0fieldidentifier or IntendedFors?

mattcieslak (approved):
This looks great. If you think we should change the current grouping behavior we should open a separate pr for that

qsiprep/reports/core.py
tsalo: Actually this fails if you enable doctests. I was originally going to enable them in this PR but there were a bunch of ones failing so I disabled that.

qsiprep/utils/grouping.py
tsalo: That's a really hard question. If you _didn't_ want to have those runs merged, how would you specify that except by not including those fields?

PR #934, opened 2025-04-05
Add tips on B1 bias correction on normed data

### LINKED ISSUE #932: Add note in documentation about N4 bias correction on prescan normalized data
We have noticed that bias correction can introduce issues on prescan normalized data. @araikes confirmed that he'd seen the same issue. We should add a note to the documentation about this (i.e., if you have prescan normalized images, use `--b1-biascorr-stage none`).

Image
Image

tsalo:
Closes #932.

- Add notes on disabling B1 bias correction on prescan-normalized data to documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -367,7 +367,7 @@ texinfo_documents = [
 linkcode_resolve = make_linkcode_resolve(
     'qsiprep',
     (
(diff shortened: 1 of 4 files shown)


mattcieslak (approved):
Thank you! I'm sure others have run into this too

PR #937, opened 2025-04-09
Bump numpy requirement

tsalo:
Closes none. I don't remember at this point what the motivation was.


tsalo:
As expected, we'll need to bump a number of other dependencies to make this work.

PR #943, opened 2025-04-28
Remove unused source_file parameter from CombineMotions

tsalo:
Closes none, but is related to #301. In #301, the source_file parameter in CombineMotions became unused, so this PR just removes the parameter to make the code more readable.

- Remove source_file from CombineMotions and clean up related functions and workflows.


mattcieslak (approved):
I remember now why src and ref file are the same image - all these images are in the same field of view

Issue #960, opened 2025-08-04
Write fieldmap in Hz to derivatives


mattcieslak:
## Summary

For benchmarking distortion on different scanners it would be nice to have fieldmaps in hz

tsalo:
In order to write out the field maps, we need to do the following:

1. Assign unique identifier to each field map, using B0FieldIdentifier, when available, or an iterator (e.g., `fmap001`) when it's not.
2. To do the first part, we need to collect all of the valid field maps at the beginning of the QSIPrep run.
3. Update the valid path patterns to support field map filenames.

tsalo:
Going to close at not planned since the fieldmap output isn't useful. The displacement map produced by #1154 should be useful enough.

Issue #949, opened 2025-06-12
New QC measures for HBCD


mattcieslak:
From the HBCD working group meeting a couple additional QC metrics 

1. How big are the distortions that are being corrected for SDC?
2. How different are the signal magnitudes in the different phase encoding directions

PR #961, opened 2025-08-04
Add fieldmap in hz to derivatives + tabular stats

mattcieslak:
Addresses #949 and #960


tsalo:
I think the main problem is that `input_image` for `fieldmap_hz_tfm` is often undefined, because only certain branches of the HMC+SDC workflows actually produce the `fieldmap_hz` image.

tsalo:
DSDTI_nofmap appears to be the only affected test.

tsalo:
I just realized that we aren't writing out the field map to the derivatives- we just have the tabular stats. Do the HBCD folks want the field map written out?

tsalo:
In that case we need to figure out how to link individual field maps with unique identifiers. SDCFlows uses an `fmapid` entity based on the B0FieldIdentifier, when available, or an integer otherwise.

Issue #962, opened 2025-08-06
Account for DRBUDDI usage in `group_dwi_scans`


tsalo:
## Summary

We need to add a grouping parameter reflecting if DRBUDDI is used. TOPUP allows various different phase encoding directions (e.g., dir-AP, dir-RL, and dir-SI are a valid combination), but DRBUDDI only allows two reverse-PEDs (dir-AP and dir-PA are solid, but don't throw dir-RL into the mix). This needs to be reflected in the groups.

Stems from https://github.com/PennLINC/qsiprep/pull/927/files#r2257931053.

PR #966, opened 2025-08-06
Remove unused parameters from `group_dwi_scans`

### LINKED ISSUE #923: `using_fsl` is hardcoded as True in `group_dwi_scans` call
## Summary

I don't really know what the impact of this bug is. It may only be a problem if a user wants to use SHORELine with reverse PE DWI scans, per @mattcieslak.

https://github.com/PennLINC/qsiprep/blob/46579b3d498e50ca722e5c4fa78fa88e4d6a4dca/qsiprep/workflows/base.py#L321-L327

tsalo:
Closes #923. This isn't a breaking change because `using_fsl` was hardcoded as True before, and `concatenate_distortion_group` only has an effect if `using_fsl` is False.

- Remove unused `using_fsl` and `concatenate_distortion_group` parameters to `group_dwi_scans` function.
- Comprehensively test the remaining parameter combinations.

### DIFF
--- a/qsiprep/utils/grouping.py
+++ b/qsiprep/utils/grouping.py
@@ -1165,41 +1165,6 @@ def group_for_eddy(all_dwi_fmap_groups):
     }
 
 
-def group_for_concatenation(all_dwi_fmap_groups):
-    """Find matched pairs of phase encoding directions that can be combined after SHORELine.
-
-    Any groups that don't have a phase encoding direction won't be correctable by SHORELine.
-
-    Parameters
-    ----------
-    all_dwi_fmap_groups : :obj:`list` of :obj:`dict`
-        A list of dictionaries describing each group of dwi files.
-
-    Returns
-    -------
-    concatenation_grouping : :obj:`dict`
-        A dictionary mapping the concatenated BIDS name of each group to the name of the
-        group that it should be concatenated with.
-    """
-    concatenation_grouping = {}
-    session_groups = _group_by_sessions(all_dwi_fmap_groups)
-    for _, dwi_fmap_groups in session_groups.items():
-        all_images = []
-        for group in dwi_fmap_groups:
-            all_images.extend(group['dwi_series'])
-        group_name = get_concatenated_bids_name(all_images)
-        # Add separate groups for non-compatible fieldmaps
-        for group in dwi_fmap_groups:
-            concatenation_grouping[group['concatenated_bids_name']] = group_name
-
-    config.loggers.workflow.info(
-        f'Found {len(concatenation_grouping)} groups of DWI series that can be concatenated:\n'
-        f'{pprint.pformat(concatenation_grouping, indent=2, width=120)}'
-    )
-
-    return concatenation_grouping
-
-
 def get_concatenated_bids_name(dwi_group):
     """Derive the output file name for a group of dwi files.
 


qsiprep/tests/test_utils_grouping.py
```diff
@@ -386,204 +374,279 @@ def test_group_dwi_scans_with_complex_b0fields(tmpdir):
...
+                },
+                {
+                    'concatenated_bids_name': 'sub-01',
+                    'dwi_series': ['sub-01_dir-PA_dwi.nii.gz'],
+                    'dwi_series_pedir': 'j',
+                    'fieldmap_info': {
+                        'rpe_series': ['sub-01_dir-AP_dwi.nii.gz'],
+                        'suffix': 'rpe_series',
+                    },
+                },
```
tsalo, line 566: It looks like all of the different PEDs aren't being combined. Also, these are conflicting concatenated BIDS names (`sub-01_dir-AP` + `sub-01_dir-PA` --> `sub-01` and `sub-01_dir-RL` + `sub-01_dir-LR` --> `sub-01`)

qsiprep/tests/test_utils_grouping.py
```diff
@@ -386,204 +374,279 @@ def test_group_dwi_scans_with_complex_b0fields(tmpdir):
...
             ],
```
tsalo, line 542: It looks like all of the different PEDs aren't being combined. Also, these are conflicting concatenated BIDS names (`sub-01_dir-AP` + `sub-01_dir-PA` --> `sub-01` and `sub-01_dir-RL` + `sub-01_dir-LR` --> `sub-01`)

qsiprep/tests/data/skeleton_simple_multiped.yml
tsalo, line 1: This skeleton will be useful when we work on #962.

qsiprep/tests/test_utils_grouping.py
mattcieslak: I think if we're not using DRBUDDI these should all go into sub-01 because TOPUP will estimate a single field from them all. I'm not sure what the bidsiest way would be to name the outputs for DRBUDDI. Maybe `sub-01_dir-APPA` and `sub-01_dir-RLLR`?

qsiprep/tests/test_utils_grouping.py
mattcieslak: yea this is unexpected

qsiprep/tests/test_utils_grouping.py
```diff
@@ -386,204 +374,279 @@ def test_group_dwi_scans_with_complex_b0fields(tmpdir):
...
+                    'fieldmap_info': {
+                        'rpe_series': ['sub-01_dir-AP_dwi.nii.gz'],
+                        'suffix': 'rpe_series',
+                    },
+                },
+            ],
+        ),
+        # Test case 3: combine_scans=False, ignore_fieldmaps=False
+        (
+            False,
```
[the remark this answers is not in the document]
mattcieslak: Also not expected behavior

mattcieslak (approved):
Great PR to get in before a comprehensive redo of the grouping code

qsiprep/tests/test_utils_grouping.py
tsalo: That's a hard one. Derivatives are still largely unchecked, but theoretically the `dir` entity corresponds to the `PhaseEncodingDirection` metadata field, meaning that the validator _should_ check that `PhaseEncodingDirection` is present when `dir` is, but it can't be since there is no PED in the combined files.

PR #967, opened 2025-08-07
Support TOPUP grouping

### LINKED ISSUE #962: Account for DRBUDDI usage in `group_dwi_scans`

tsalo:
Closes #962.

- Group all scans with compatible PEDs when not using DRBUDDI.


tsalo:
It looks like I need to fix downstream functions- not just group_by_warpspace.

tsalo:
Closing in favor of #1092.

PR #1000, opened 2025-11-12
Allow configuration of `--no-csf` in SynthStrip

### LINKED ISSUE #954: Spatial normalization targeting dura(?) instead of brain in presence of atrophy
## Summary

In several participants exhibiting a decent amount of atrophy, the spatial normalization step isn't really adjusting the outer cortical regions and is instead appears to match the dura of the patient to the edge of the brain of the template. Registration is great in many parts of the brain (e.g., genu of corpus callosum), but not near any edge of cortex that has receded from the skull. In this example case, the see the superior part of the frontal cortex on the x=13 and x=-14 slices

Image
Image

## Additional details

- QSIPrep version: 0.21.5.dev0+g36b93fe.d20240504
- Docker version: pennbbl/qsiprep:0.21.4
- Singularity version: apptainer/1.4.1

Process older participants, some of whom have a decent amount of atrophy.

I expected better normalization in the frontal cortex.

### What actually happened?

Here's my best guess at the source of the issue. QSIprep makes a brain mask with `mri_synthstrip`, which is _amazing_ at what it's designed to do. However, the mask that `mri_synthstrip` creates tends to include dura, which is especially noticeable in brains that have lots of atrophy (e.g., see their T1w example: https://surfer.nmr.mgh.harvard.edu/docs/synthstrip/resources/examples.png), although this can be controlled with the `--no-csf` flag. In contrast, the template brain masks are tighter. 

Image

I suspect that the call to `antsRegistration` does a great job ensuring that the edges of the fixed and moving images are aligned, but that ends up being a problem, because the "edges" are different kinds of tissue, brain and dura respectively. 

It is my impression that an ANTs tool like `antsBrainExtraction.sh` targets something different than `mri_synthstrip`, which is a tighter mask that excludes dura (and csf), more comparable to what is available for template images on templateflow. 

Even if that's the source of the issue, I'm not really sure what would be the solution. Maybe call `mri_synthstrip` with the `--no-csf` flag? Alternatively, replace the template mask with one created by a default call to `mri_synthstrip`? 

I have lots of these examples, so if it's agreed that A) this is an issue, and B) a likely source, then I'd be happy to experiment with solutions. 

## Reproducing the bug

FWIW,

```
args=['qsiprep', '--fs-license-file', '/opt/qsiprep_app/license.txt', '--output-resolution', '1.7', '--hmc_model', 'eddy', '--eddy-config', '/opt/qsiprep_app/eddy_params.json', '--unringing-method', 'mrdegibbs', '--denoise-method', 'patch2self', '--notrack', '--skip-bids-validation', '--work-dir', '/tmp/tmpg7dxtyvq', '--nthreads', '16', '--mem_mb', '32000', '/tmp/tmpol1gpc2z', '/tmp/tmpe2fisj5y', 'participant']
```

psadil:

Allow configuration of the `--no-csf` option in the SynthStrip workflow ~~and by default set it to True (currently implies False)~~. Keep the currently implied default (False).

Closes #954 

---

> [!NOTE]
> Adds a `--no-csf` CLI flag and `workflow.no_csf` config, plumbing the option into SynthStrip across anatomical, DWI, and fieldmap workflows.
> 
> - **CLI**:
>   - Import `BooleanOptionalAction` and add `--no-csf` flag in `qsiprep/cli/parser.py` to control SynthStrip CSF removal.
> - **Config**:
>   - Introduce `workflow.no_csf` boolean in `qsiprep/config.py`.
> - **Workflows**:
>   - Update `init_synthstrip_wf` to accept `no_csf` and pass it to `FixHeaderSynthStrip`/`MockSynthStrip`.
>   - Propagate `no_csf=config.workflow.no_csf` in:
>     - Anatomical volume workflows (`init_anat_preproc_wf`, T2w preprocessing).
>     - DWI reference workflow (`init_dwi_reference_wf`).
>     - PEPOLAR fieldmap workflow (`init_extended_pepolar_report_wf`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93e66300baba920a7bf64d39781b4ce19f351c6c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update && \
 COPY . /src/qsiprep
 RUN python -m build /src/qsiprep
 
-FROM pennlinc/qsiprep_build:24.10.0
+FROM pennlinc/qsiprep_build:26.1.7
 
 # Install qsiprep wheel
 COPY --from=wheelstage /src/qsiprep/dist/*.whl .


mattcieslak (dismissed):

psadil:
@tsalo , @mattcieslak , sorry, but I'm hitting a small snag. I only just realized that the version of `mri_synthstrip` bundled into the `qsiprep_build` image [doesn't have the `--no-csf` option](https://github.com/freesurfer/freesurfer/blob/115e4660f11e66f3a4ed354f00bc01a3c8440dd7/mri_synthstrip/mri_synthstrip#L22-L28). It was introduced in [mri_synthstrip 1.2](https://hub.docker.com/r/freesurfer/synthstrip). What would be the preferred way to handle that? Options I see include (ordered roughly from smallest to largest change)

- bundle [the nocsf model](https://github.com/freesurfer/freesurfer/blob/v8.1.0/mri_synthstrip/synthstrip.nocsf.1.pt)
  - [the `--no-csf` option just selects this model](https://github.com/freesurfer/freesurfer/blob/f30dfed066e8357755d28b555c266153b697bd0b/mri_synthstrip/mri_synthstrip#L252-L256)
- have `qsiprep_build` pull `mri_synthstrip` from the [freesurfer/synthstrip image](https://hub.docker.com/r/freesurfer/synthstrip), instead of the freesurfer/freesurfer one
- update the FreeSurfer version in `qsiprep_build`
  - e.g., [v7.3.2 would do it](https://github.com/freesurfer/freesurfer/blob/635427507bc692c2393393f505d0ee431f68747a/mri_synthstrip/mri_synthstrip#L30)

PR #1028, opened 2026-03-04
Adopt NiPreps-style packaging

tsalo:
Closes none.

- Move FSL and Connectome Workbench requirements into Dockerfile.
- Add Dockerfile.base with non-Python, non-conda requirements. This still builds off of the package-wise Docker images in `qsiprep_build`, except for the main Dockerfile there.
- Add pixi.lock file.
- Add pixi-lock GitHub Action.
- Stop using `pkg_resources`, which is no longer supported. Use `importlib.resources` instead.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,9 +44,7 @@ The main features of this software are
   2. Automatically generated preprocessing pipelines that correctly group, distortion correct,
      motion correct, denoise, coregister and resample your scans, producing visual reports and
      QC metrics.
-  3. A system for running state-of-the-art reconstruction pipelines that include algorithms
-     from Dipy_, MRTrix_, `DSI Studio`_  and others.
-  4. A novel motion correction algorithm that works on DSI and random q-space sampling schemes
+  3. A novel motion correction algorithm that works on DSI and random q-space sampling schemes
 
 .. image:: https://github.com/PennLINC/qsiprep/raw/master/docs/_static/workflow_full.png
 
@@ -81,5 +79,5 @@ Note
 
 The *QSIPrep* pipeline uses much of the code from *fMRIPrep*.
 It is critical to note that the similarities in the code
-**do not imply that the authors of QSIPrep in any way endorse or support this code or its
+**do not imply that the authors of fMRIPrep in any way endorse or support this code or its
 pipelines**.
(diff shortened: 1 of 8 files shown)


.cursor/rules/cursorenv.mdc
mattcieslak, line 1: I'm not sure we need this file. it might be a good .gitignore candidate

qsiprep/tests/test_cli.py
```diff
@@ -446,34 +450,80 @@ def test_multi_t1w(data_dir, output_dir, working_dir):
 
     This tests the following features:
     - freesurfer's robust template
```
mattcieslak, line 452: I don't think we actually use freesurfer's robust template anymore

qsiprep/tests/test_cli.py
```diff
@@ -446,34 +450,80 @@ def test_multi_t1w(data_dir, output_dir, working_dir):
...
+        writable_bids_dir,
         out_dir,
         'participant',
         f'-w={work_dir}',
-        '--b1-biascorrect-stage=none',
-        '--hmc-model=none',
-        '--b0-motion-corr-to=first',
+        f'--eddy-config={eddy_config}',
+        '--denoise-method=none',
+        '--sloppy',
```
mattcieslak, line 506: Why the parameters change here?

tsalo: I don't believe `multi_t1w` was a functioning pytest previously, so I asked Codex to translate it and the other non-pytest test into Python in ccc1ff7cd948fed6d6ee66548ba650c94d3487e2.

qsiprep/tests/test_cli.py
tsalo: I think a lot of the docstrings for the integration tests are outdated/incorrect. We should probably update them at some point.

.cursor/rules/cursorenv.mdc
tsalo: Good point! I'll remove it.

tsalo:
@mattcieslak since you've approved, I'll merge this once CI passes. That alright?

mattcieslak:
Sounds great! I think we may want to switch the main branch to `main` at some point too, since all the other apps use it

Issue #1067, opened 2026-06-23
QSIPrep 26.0.0 Singularity image: eddy_cuda10.2 missing and DSI Studio fails with GLIBC_2.38 under --nv


bkossows:
I am trying to run `pennlinc/qsiprep:26.0.0` converted to a Singularity image with GPU support:

```bash
singularity run --nv qsiprep-26.0.0.simg ...
```

Two issues appear.

First, QSIPrep selects CUDA eddy but looks specifically for `eddy_cuda10.2`:

```text
Using CUDA and 1 threads in eddy
Cannot run QSIPrep. Missing dependencies:
* eddy_cuda10.2 (Interface: ExtendedEddy)
```

However, the image contains:

```text
/app/.pixi/envs/qsiprep/bin/eddy_cpu
/app/.pixi/envs/qsiprep/bin/eddy_cuda11.0
```

A workaround is to bind a wrapper/symlink named `eddy_cuda10.2` pointing to `eddy_cuda11.0`.

Second, after fixing the eddy name issue, DSI Studio fails when running with `--nv`:

```text
dsi_studio: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found
(required by /.singularity.d/libs/libGLdispatch.so.0)

dsi_studio: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found
(required by /.singularity.d/libs/libGLX.so.0)
```

This happens for commands such as:

```text
dsi_studio --action=src ...
```

The container uses `/opt/dsi-studio/dsi_studio`, so this does not seem to be caused by a host-side `dsi_studio` shadowing the container binary.

Environment:

```text
QSIPrep image: pennlinc/qsiprep:26.0.0 converted to Singularity .simg
Runtime: Singularity with --nv
Command includes: --output-resolution 1.5 --nthreads 16 --omp-nthreads 8 --skip-bids-validation
```

tsalo:
Can you try using the unstable Docker image? I believe I fixed this bug recently, but we haven't made a new release since that PR was merged.

samuelholly:
Hello, I ran into the same problem with `eddy_cuda` today. I can confirm that the bug is resolved in the unstable version, eddy runs fine.

I tried it with Docker on Windows 11.

Hope this helps.

Best regards

Samuel

tsalo:
[...]

PR #1073, opened 2026-07-23
Configure shared libraries for DSIStudio commands at runtime

tsalo:
I believe this should address #1067. 

- Exclude host libraries added to Apptainer when using `--nv` in DSIStudio interface.


tsalo:
Per convo with @mattcieslak, I'm going to merge as-is and revert if there are problems.

PR #1080, opened 2026-08-10
Address inheritance failures

### LINKED ISSUE #685: The inheritance principle is not completely supported
I haven't done anything near a complete inventory of where metadata is loaded in QSIPrep, but in at least one location ([`get_metadata_for_nifti`](https://github.com/PennLINC/qsiprep/blob/527ef590f1c2bd77c8172cb9bf5e0f4cc09ea7b9/qsiprep/interfaces/bids.py#L457)), inherited metadata is not support.

I think the best solution is to use a `BIDSLayout` to load metadata instead of directly finding the (potentially multiple) JSONs associated with a NIFTI. This might be easier if we implement the Nipreps `Config` object.

### LINKED ISSUE #990: bval and bvec file identification does not account for inheritance principle
> Found another case, where bval and bvec files are grabbed based on the DWI filename. I'm not sure where in the code this happens (other than that it's in the merge workflow), but that seems to be the case.
> 
> 
> 
> Per chat with @mattcieslak, bvals should vary by run, _but_ inheritance should definitely apply to magnitude and phase images, which come from the same run. So you'd typically have `sub-X_dwi.bval` for `sub-X_part-mag_dwi.nii.gz` _and_ `sub-X_part-phase_dwi.nii.gz`. 

 _Originally posted by @tsalo in [#685](https://github.com/PennLINC/qsiprep/issues/685#issuecomment-1921462739)_

tsalo:
Closes #685 and closes #990.

Removed from qsiprep/interfaces/fmap.py:
- CRITICAL_KEYS
- acqp_lines
- _merge_metadata
- read_nifti_sidecar
- load_epi_dwi_fieldmaps
- get_distortion_grouping
- eddy_inputs_from_dwi_files
- get_topup_inputs_from

All but the last were byte-identical duplicates of definitions already in epi_fmap.py; fmap.py now imports them from there. get_topup_inputs_from was dead and is deleted outright.

Added to qsiprep/utils/bids.py:

- find_bids_root(path)
- _parse_bids_name(path)
- _inheritance_levels(path)
- find_associated_files(path, extension)
- load_sidecar(path)
- find_bval(path)
- find_bvec(path)


tsalo:
I had Opus 4.8 review the PR:

All 44 new tests pass. Now let me examine the remaining diffs (eddy, tortoise, dwi_merge) and read the test files to assess coverage and look for edge cases.

Let me read the main test file to assess coverage of the new inheritance logic.

Let me look at the epi_fmap.py new `read_nifti_sidecar` and its call sites in context, plus check for circular import safety.

Let me check whether any existing (non-new) tests or code still import the removed names from `fmap`.

Let me check for now-unused imports left behind in the edited files (a common refactor oversight).

Let me check whether other spots in the codebase still use the old "beside-the-file" assumption for bval/bvec/json, to gauge completeness of the fix.

`dwi_merge.py:335-336` still uses the beside-file assumption. Let me check whether that path handles the same complex/inheritance case.

I have a thorough picture now. Let me verify one last thing — that `generate_bids_skeleton` doesn't emit per-part JSONs that would collide with the shared-file scenario (confirming the targeted complex-data path is safe), by checking what the merge test actually exercises end-to-end.

I've completed a thorough review. Here's my assessment.

## Review: `fix-inheritance` PR

**What it does:** Fixes qsiprep's handling of the BIDS *inheritance principle* for DWI sidecars and `.bval`/`.bvec` files — the long-standing limitation (issues #685, #990) where complex-valued acquisitions had to duplicate `part-mag`/`part-phase` gradient and metadata files instead of sharing a single `sub-<label>_dwi.bval`. It also deduplicates a set of functions that had drifted between `fmap.py` and `epi_fmap.py`.

**Verification I ran (lincapps env, WSL):**
- All 44 new tests pass (`test_utils_bids`, `test_interfaces_images`, `test_interfaces_fmap`, `test_interfaces_epi_fmap`).
- `ruff check` clean on all changed files (no leftover unused imports from the refactor).
- Import smoke test passes — no circular import from the new `fmap.py → epi_fmap.py` dependency.
- No stale references to the removed `fmap.py` functions; `get_topup_inputs_from` was genuinely dead (zero callers).

### Strengths
- The new [`[qsiprep/utils/bids.py](https://claude.ai/epitaxy/qsiprep/utils/bids.py)`](qsiprep/utils/bids.py) helpers are clean, well-documented, and correctly implement the inheritance semantics: shallowest-first ordering with nearest-wins merge, subset-entity matching, per-directory ambiguity detection, and a sensible fallback (files outside a dataset only match siblings — important for copied working-dir files).
- The fix is applied at the right boundary. `ConformDwi` and `load_epi_dwi_fieldmaps` are where gradients first enter from BIDS; downstream (`dwi_merge.py:335`) only builds *output* names from already-resolved inputs, so it correctly needed no change.
- Test coverage is genuinely good — root/subject/session inheritance, conflicting entities, wrong suffix, extra-entity specificity, nearest-wins, outside-dataset fallback, and the exact complex-data scenario are all covered.
- Nice cleanup: removes the now-obsolete docs workaround and the duplicated implementations.

### Minor findings (non-blocking)

1. **`find_associated_files` raises `ValueError` on two applicable files in one directory** ([[bids.py:189](https://claude.ai/epitaxy/qsiprep/utils/bids.py:189)](qsiprep/utils/bids.py:189)). This is spec-correct, but it's a behavior change from "silently use the beside-file." A dataset with both `sub-01_dwi.json` and `sub-01_part-mag_dwi.json` colocated (invalid BIDS, but tolerated before) will now hard-error rather than proceed. Worth being an intentional choice — the error message is clear and points at the spec, so I'd keep it, just flagging the change.

2. **Pre-existing latent `TypeError` surfaces more paths now:** `get_distortion_grouping` does `acqp_line = spec_line % spec['TotalReadoutTime']` ([[epi_fmap.py:166](https://claude.ai/epitaxy/qsiprep/interfaces/epi_fmap.py:166)](qsiprep/interfaces/epi_fmap.py:166)); the new `read_nifti_sidecar` returns `TotalReadoutTime=None` when absent, so a sidecar found via inheritance but missing `TotalReadoutTime` would raise a `TypeError` on the `%` format rather than a clear message. This isn't introduced by the PR (old code had the same gap), but since inheritance now resolves *more* sidecars, it's slightly more reachable. A targeted "missing TotalReadoutTime" error would be friendlier — optional.

Overall this is a high-quality, well-tested PR that correctly solves the stated problem and leaves the code cleaner than it found it. I'd approve it. The two notes above are polish, not blockers.

PR #1061, opened 2026-06-11
Refactor scan grouping

tsalo:
Keeping as a draft for now.

## Open issues

- DRBUDDI RPE.

## Summary

Reorganizes QSIPrep's scan-grouping and field-map-selection code into SDCFlows-style value objects, without changing behavior. The previously monolithic qsiprep/utils/grouping.py is split into a clear pipeline and shrinks from 1725 → 1055 lines.

- New qsiprep/utils/fieldmaps.py (+217) mirroring sdcflows.fieldmaps: EstimatorType, MODALITIES, FieldmapFile (sibling discovery), and FieldmapEstimation (method inference, bids_id, and to_fieldmap_info() — the single place per-suffix field-map knowledge now lives).
- Slimmed grouping.py: DwiSeries → _final_distortion_groups → find_estimators (B0FieldIdentifier → IntendedFor →
PE-heuristic) → _serialize_estimation_groups + build_fmap_application_groups, with group_dwi_scans reduced to a thin orchestrator.
- base.py adapter builds fieldmap_info via FieldmapEstimation.to_fieldmap_info() instead of re-deriving suffix typing.
- Deleted the post-hoc refine/remap dance (refine_distortion_groups, _remap_*) and the duplicated per-suffix helpers (classify_fmap_files, get_highest_priority_fieldmap, FMAP_PRIORITY, the legacy build_distortion_groups/build_fmap_estimation_groups and their helpers) — ~670 lines net removed.

## Why it's safe (behavior preservation)

- Adds a golden/characterization suite (test_grouping_golden.py, 240 snapshots) frozen from the pre-refactor
group-dwi-scans baseline, covering every fixture × combine_scans/ignore_fieldmaps/estimate_per_axis combo, including error cases. The new code reproduces it exactly and deterministically.
- For the trickiest edge cases (same-physical-signature multi-run DWIs with per-run field-map links via
B0FieldSource/IntendedFor), the four-dict output was verified byte-identical to the pre-refactor baseline across all 8
flag combos, locked by two dedicated fixtures.
- Also fixed pre-existing nondeterminism in three grouping error messages (Python set reprs reordered between runs), so the suite is stable.

## Test plan

- [ ] pytest qsiprep/tests/test_grouping_golden.py qsiprep/tests/test_utils_grouping.py qsiprep/tests/test_utils_fieldmaps.py — all green.
[...]

Issue #965, opened 2025-08-06
Clarify what `--ignore fieldmaps` means


tsalo:
In today's call, we clarified that ignoring "fieldmaps" meant skipping distortion correction completely, rather than just ignoring fieldmaps (and still potentially using reverse-PE DWI scans for distortion correction).

PR #1092, opened 2026-08-14
Redo the scan grouping so it uses modern BIDS and is easier to understand

### LINKED ISSUE #273: What to do about different FoV's
# Background

FSL-style bvec files specify gradient directions in the reference frame of the image axes. The affine of the corresponding 4D image can be used to get the gradient direction in scanner coordinates. 

Often times the sampling scheme is broken up into multiple runs. When runs are acquired with the exact same FoV, it's no problem to concatenate the images and bvec files. However, in the real world the FoVs can change when scanner operators move the FoV. 

# Problem:

Images (and their gradient directions) need to be compared to one another at numerous points. If the x, y, z dimensions are the same, we can take the 3d matrices and *concatenate* them in the 4th dimension. This does not involve any data interpolation. 

A problem arises when these same-shaped 3D images occupy different coordinates in space. Even though the data matrices can be concatenated, the mapping from a voxel coordinates to scanner coordinates will then change from one image to another. This may or may not be a big deal depending on how the mapping to scanner coordinates changes between images. The 4D nifti file of concatenated images also has **only one** official affine. So the bvecs need to match this image axes defined by this affine. 

## Case 1: Image FoVs are shifted and not rotated

This will have no effect on the bvec files because the image axis is not rotated between the scans. Any shifts in x,y,z will appear as head motion.

Will this cause any issues for distortion correction? The slices are still in the same plane, but the images are in different locations relative to the scanner's isocenter. 

## Case 2: Images are rotated

If two images are acquired with FoVs rotated relative to one another, the bvec directions need to be adjusted so they are pointing in the correct direction **relative to the official affine in the concatenated image**.

Does this cause issues for distortion correction?

## Current approach

To concatenate images with the same i, j, k dimensions we do the following:

1. Use mrtrix to convert the fsl bvec files to mrtrix b files. b files are in RAS+ scanner space
2. Copy the affine from the first image into the headers of all subsequent images
3. Use mrtrix again to `mrconvert` the b files into fsl bvecs corresponding to the header-adjusted nifti file

After 3, all images will have the same affine and a bvec file corresponding to it. I think you can concatenate these. The differences in head position will be interpreted as head motion.

### Issues related to this problem

#106, #153, #218, #225

### LINKED ISSUE #597: Dimension mismatch when using --prefer_dedicated_fmaps
I'm getting a dimension mismatch error when I try to use `--prefer_dedicated_fmaps`. 

Here is my call to qsiprep:
```
singularity run --cleanenv /scratch/users/ethanroy/singularity_images/qsiprep-0.18.1.sif \
/scratch/groups/jyeatman/ethan/input_test_fmap_reorg \
/scratch/groups/jyeatman/ethan/input_test_fmap_reorg/derivatives participant --participant-label sub-LD4138ses3 \
--prefer_dedicated_fmaps \
-w /scratch/groups/jyeatman/ethan/work \
--output-resolution 1.5

```

And here is the stack trace:
```
RuntimeError: Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 527, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 645, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 771, in _run_command
    raise NodeExecutionError(msg)
nipype.pipeline.engine.nodes.NodeExecutionError: Exception raised while executing Node fmap2ref_apply.

Cmdline:
	antsApplyTransforms --default-value 0 --dimensionality 3 --float 1 --input /scratch/groups/jyeatman/ethan/work/qsiprep_wf/single_subject_LD4138ses3_wf/dwi_preproc_acq_60_wf/hmc_sdc_wf/sdc_wf/fmap_wf/applymsk/sub-LD4138ses3_acq-60_fieldmap_ras_rad_unwrapped_hz_filt_demean_maths_masked.nii.gz --interpolation BSpline --output sub-LD4138ses3_acq-60_fieldmap_ras_rad_unwrapped_hz_filt_demean_maths_masked_trans.nii.gz --reference-image /scratch/groups/jyeatman/ethan/work/qsiprep_wf/single_subject_LD4138ses3_wf/dwi_preproc_acq_60_wf/hmc_sdc_wf/b0_ref_to_lps/topup_imain_LPS.nii.gz --transform /scratch/groups/jyeatman/ethan/work/qsiprep_wf/single_subject_LD4138ses3_wf/dwi_preproc_acq_60_wf/hmc_sdc_wf/sdc_wf/sdc_unwarp_wf/fmap2ref_reg/transformComposite.h5
Stdout:

Stderr:

Traceback:
	Traceback (most recent call last):
	  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 398, in run
	    runtime = self._post_run_hook(runtime)
	  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/niworkflows/interfaces/registration.py", line 106, in _post_run_hook
	    return super(ANTSApplyTransformsRPT, self)._post_run_hook(runtime)
	  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/interfaces/mixins/reporting.py", line 50, in _post_run_hook
	    self._generate_report()
	  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/niworkflows/interfaces/report_base.py", line 61, in _generate_report
	    cuts = cuts_from_bbox(mask_nii, cuts=n_cuts)
	  File "/usr/local/miniconda/lib/python3.8/site-packages/qsiprep/niworkflows/viz/utils.py", line 215, in cuts_from_bbox
	    ras_coords.append(apply_affine(mask_nii.affine, cross).tolist())
	  File "/usr/local/miniconda/lib/python3.8/site-packages/nibabel/affines.py", line 97, in apply_affine
	    pts = pts @ rzs.T + trans[None, :]
	ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 3 is different from 4)

```

### LINKED ISSUE #658: Intendedfor fmap entry update to accept "bids::sub-xx"
Hi experts, 
I am a little bit confused and would appreciate any help.
We convert our data with bidscoiner into BIDS. Fieldmaps can be automatically assigned to (for example) DWI-images using the "Intendedfor" tag. As stated in the latest [bids-specifications](https://bids-specification.readthedocs.io/en/latest/modality-specific-files/magnetic-resonance-imaging-data.html#using-intendedfor-metadata), this would be done as followed:
"IntendedFor": [
        "bids::sub-01/ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
        "bids::sub-01/ses-pre/func/sub-01_ses-pre_task-motor_run-2_bold.nii.gz"]}   
The latest bids-validator is happy with this notation !!

However, qsiprepr (19.1) - more precisely the bids-validator provided in that container - doesn't find the referenced images. It took me a while to figure out that the following notation is needed:

"IntendedFor": [
        "ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
        "ses-pre/func/sub-01_ses-pre_task-motor_run-2_bold.nii.gz"]}   

So, "bids::sub-xx" should not be used. 

Is there a way to fix this problem? Is there anyone else out there having this problem?

Thanks for your help

Karl

### LINKED ISSUE #919: Support B0FieldSource and B0FieldIdentifier metadata fields
## Summary

B0FieldSource and B0FieldIdentifier are the preferred way to specify what files should be used for susceptibility distortion correction in BIDS. Currently, QSIPrep only supports IntendedFor with relative paths.

This will allow QSIPrep to support more custom distortion correction configurations, since users can say exactly which files they want to use for SDC.

Related to #658.

### LINKED ISSUE #921: `get_concatenated_bids_name` may be overly aggressive
## Summary

This is something I noticed while looking at #920. I don't think it's the bug _causing_ #920, but it's related. The function `get_concatenated_bids_name` will remove overlapping entities within a list of BIDS files. For example:

```python
>>> get_concatenated_bids_name([
...    '/data/sub-1/dwi/sub-1_dir-AP_run-1_dwi.nii.gz',
...    '/data/sub-1/dwi/sub-1_dir-AP_run-2_dwi.nii.gz',
...    '/data/sub-1/dwi/sub-1_dir-AP_run-3_dwi.nii.gz',
...    '/data/sub-1/dwi/sub-1_dir-AP_run-4_dwi.nii.gz'
... ])
'sub-1_dir-AP'
```

However, in the scenario where `run-1` and `run-2` use one field map and `run-3` and `run-4` use another, `get_concatenated_bids_name` will be called twice:

```python
>>> get_concatenated_bids_name([
...    '/data/sub-1/dwi/sub-1_dir-AP_run-1_dwi.nii.gz',
...    '/data/sub-1/dwi/sub-1_dir-AP_run-2_dwi.nii.gz',
... ])
'sub-1_dir-AP'

>>> get_concatenated_bids_name([
...    '/data/sub-1/dwi/sub-1_dir-AP_run-3_dwi.nii.gz',
...    '/data/sub-1/dwi/sub-1_dir-AP_run-4_dwi.nii.gz',
... ])
'sub-1_dir-AP'
```

This results in duplicate concatenated BIDS names for the two distortion groups.

In #920's case, each distortion group only had one run, so that isn't the problem.

### LINKED ISSUE #930: SyN-based susceptibility distortion correction not carried out when using --use-syn-sdc
## Summary
The --use-syn-sdc flag is supposed to invoke SyN-based susceptibility distortion correction, however the flag has no effect at all. 

## Additional details
Preprocessing was performed using *QSIPrep* 1.0.1.dev0+gee9aa2e.d20250115 [@cieslak2021qsiprep]

Run QsiPrep with SyN-based susceptibility correction

That SyN-based susceptibility correction was performed.
The work-around is to use the flag --force-syn,
but according to the documentation that flag should be used to use SyN-based susceptibility correction *in addition to* fieldmap based correction.

### What actually happened?
Susceptibility correction was not used at all.
See also the expected behavior as described here:
https://github.com/PennLINC/qsiprep/blob/fc899458ed1b23922e14a013f674c4d6ac029264/qsiprep/workflows/fieldmap/base.py#L25

## Reproducing the bug
To reproduce the bug just run QsiPrep on a dataset that does not have fieldmaps or reverse phase encoding available with the option --use-syn-sdc. To speed up testing --boilerplate-only can be specified, because from the boilerplate text it is already clear whether syn-sdc is included in the workflow.

```
```
-->

### LINKED ISSUE #962: Account for DRBUDDI usage in `group_dwi_scans`

### LINKED ISSUE #963: Enumerate effects of combinations of MultipartID and B0Field* fields
This is something that came up in a call today with @mattcieslak and @smeisler. We want to write documentation (probably with diagrams) that shows how (1) MultipartID, (2) B0FieldSource/B0FieldIdentifier, and (3) QSIPrep parameters will impact distortion correction and run concatenation, including `--ignore fieldmaps`, `--separate-all-dwis`, `--distortion-group-merge`, and `--pepolar-method`.

### LINKED ISSUE #964: EPI field map collection is too aggressive- may ignore IntendedFor field
## Summary

In one of my tests in #927, I have dir-AP_epi and dir-PA_epi field maps, but the IntendedFor only points to reverse PED DWI runs (i.e., dir-AP_epi --> dir-PA_dwi, dir-PA_epi --> dir-AP_dwi), and yet `group_dwi_scans` selects both dir-AP_epi and dir-PA_epi as field maps for the dir-PA DWI scan.

This stems from https://github.com/PennLINC/qsiprep/pull/927/files#r2257900827.

### LINKED ISSUE #1036: Use tolerances for affine matrix differences to allow for floating point errors
## Summary

`nilearn.image.concat_imgs` appears to be overly conservative in how it compares affines to determine if they can be concatenated. Currently, we tell users to overwrite their affines manually, which is suboptimal. Instead, we could "look at the zooms, offsets and angles and set a threshold for each. The zooms had better be within eps of each other, same with the offsets. The angles we can calculate differences in degrees and go from there" (@mattcieslak's idea). If DWI files are compatible according to these thresholds, then we can create a modified version of each image with the affine from the first image, then pass them all into `concat_imgs`.

Stems from https://github.com/PennLINC/qsiprep/issues/225#issuecomment-3883223351.

### LINKED ISSUE #1090: Check ShimSettings when building groups
Also incorporate this into the recommendation CLI.

mattcieslak:
The qsiprep scan grouping has been a pain for years, and needs an overhaul so it supports modern bids features like b0 field identifiers.

Supersedes #967 and #1061.

Closes #1090 
Closes #1036 
Closes #964 
Closes #963 
Closes #962 
Closes #930
Closes #921 
Closes #919 
Closes #658 
Closes #597 
Closes #273


qsiprep/tests/test_grouping_inference.py
```diff
@@ -0,0 +1,375 @@
...
+
+def test_bidsuri_intendedfor(tmp_path):
+    """bids:: URIs resolve identically to relative paths."""
+    grouping = load_scenario('bidsuri_intendedfor', tmp_path)
+    (estimation,) = grouping.estimations.values()
+    assert estimation.provenance is Provenance.TRANSLATED
+    assert sorted(basenames(estimation.sources)) == [
+        'sub-01_dir-AP_dwi.nii.gz',
+        'sub-01_dir-PA_epi.nii.gz',
+    ]
```
tsalo, line 98: The basenames would be the same even if the BIDS URIs aren't resolved correctly, right? Can you have it check that the files exist?

qsiprep/grouping/metadata.py
```diff
@@ -0,0 +1,327 @@
...
+    ----------
+    layout : :class:`bids.BIDSLayout`
+        Used for sidecar reading (``get_metadata`` applies BIDS inheritance)
+        and for discovering fmap/anat files when ``subject_data`` lacks them.
+    subject_data : dict
+        As produced by :func:`qsiprep.utils.bids.collect_data`: at minimum a
+        ``'dwi'`` key listing this subject's DWI files; optional ``'fmap'``,
+        ``'t1w'``, and ``'t2w'`` keys.
+    ignore_fieldmaps : bool
+        Skip fmap/ indexing entirely (``--ignore fieldmaps``). The DWI-based
```
tsalo, line 300: This seems to change the behavior of `--ignore fieldmaps`. The current documentation says `“fieldmaps” will completely disable susceptibility distortion correction, whether using field maps or reverse phase-encoded dMRI runs.`, based on #965.

qsiprep/grouping/__main__.py
```diff
@@ -0,0 +1,129 @@
...
+"""
+
+import argparse
+import sys
+
+from . import build_dwi_grouping, describe_processing, report_text
+from .validation import BACKENDS
+
+
+def _build_parser():
```
tsalo: We normally put any CLIs in the `cli` module.

qsiprep/grouping/models.py
```diff
@@ -0,0 +1,470 @@
...
+from collections.abc import Mapping
+from types import MappingProxyType
+from typing import Any
+
+import numpy as np
+
+#: Reserved prefix for identifiers generated by qsiprep. ``+`` cannot appear
+#: in a BIDS filename entity and is not expected in curated identifiers, so a
+#: curated value starting with this prefix is treated as an error.
+AUTO_PREFIX = 'auto+'
```
tsalo, line 39: This is only used in inference.py, so why not move it there?

qsiprep/grouping/models.py
```diff
@@ -0,0 +1,470 @@
...
+
+#: Seconds; TotalReadoutTime values closer than this are considered equal.
+READOUT_TOLERANCE = 1e-4
```
tsalo, line 42: This is only used in metadata.py.

grouping_integration_plan.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+## Invariants — do not violate during integration
+
+1. **The model describes data, never tool arrangements.** No `rpe_series`-style
+   synthetic vocabulary enters `qsiprep/grouping/`. Tool-specific shapes are
+   produced by adapters that consume the model.
+2. **Curated > translated > forced > inferred**, and every decision carries its
+   `Provenance`. A real fieldmap always beats fieldmap-less methods.
+3. **Estimation membership ≠ concatenation membership** (borrowing). An
```
tsalo: I don't think this matches what we've discussed in the past (e.g., https://pennbbl.slack.com/archives/C07GANDUTGA/p1773244511125539, which I tried to follow in #967). Namely, in the past we decided that concatenation groups are composed of field map estimation groups.

PR #1105, opened 2026-08-25
Introduce the "workflow compiler"

mattcieslak:
We moved the workflow compiler out of qsiprep and into qsiplan. From there a quick pip install can help users curate and plan an analysis for their BIDS data. It removes a lot of graphics and testing out of the qsiprep package


qsiprep/cli/group.py
```diff
@@ -35,12 +44,31 @@ def _build_parser():
...
+    )
     parser.add_argument(
         '--backend',
         nargs='+',
         choices=BACKENDS,
-        default=list(BACKENDS),
-        help='Backend(s) to preview. Default: all three.',
+        default=None,
+        help='DEPRECATED: use --hmc-method/--sdc-method. Legacy backend '
+        'name(s) to preview.',
```
tsalo: I don't think you need to deprecate `--backend` since it was never released. I'd drop it completely and deprecate the old flags (`--hmc-model`, `--pepolar-method`).

tsalo: Actually, since the grouping CLI itself is unreleased, there's no reason to deprecate anything in this argument parser.

qsiprep/cli/parser.py
```diff
@@ -716,26 +790,46 @@ def _bids_filter(value, parser):
...
+        '"eddy" means `--hmc-method eddy`; "tortoise" means `--hmc-method '
+        'tortoise`; "3dSHORE", "tensor" and "none" mean `--hmc-method '
+        'shoreline` with the matching --shoreline-model.',
+    )
+    g_moco.add_argument(
+        '--shoreline-model',
+        action='store',
+        default=None,
+        choices=['3dshore', 'tensor', 'none'],
+        help='signal model SHORELine uses to predict motion-correction target '
```
tsalo, line 844: Do you want to deprecate this (even though you're adding it now) so it's easier to drop shoreline completely in the next major release?

qsiprep/workflows/base.py
```diff
@@ -74,24 +73,21 @@
 from .dwi.util import get_source_file
 
 
-def _t2w_sdc_backend_enabled():
-    """Whether the selected backend has a stage that can consume a T2w for SDC.
+def _t2w_sdc_enabled(selection):
```
tsalo: Since you're touching this, I'd move it to `utils`.

qsiprep/workflows/base.py
```diff
@@ -74,24 +73,21 @@
...
     """
-    return (
-        'drbuddi' in (config.workflow.pepolar_method or '').lower()
-        or config.workflow.hmc_model == 'tortoise'
-    )
+    return SdcTool.DRBUDDI in selection.pepolar_tools or selection.hmc is HmcMethod.TORTOISE
 
 
-def _t2w_available_for_sdc(subject_data):
+def _t2w_available_for_sdc(subject_data, selection):
```
tsalo: Since you're touching this, I'd move it to `utils` and just pass in any variables from the module namespace to the function as arguments.

qsiprep/config.py
```diff
@@ -591,10 +591,16 @@ class workflow(_Config):
     """Remove the mean from fieldmaps."""
     force_syn = None
     """Run *fieldmap-less* susceptibility-derived distortions estimation."""
+    force_t2wreg = None
```
tsalo: Worth considering dropping individual `--force-[X]` parameters in favor of `--force [X] [Y] ...` parameter.

PR #1106, opened 2026-08-26
Support gradient unwarping with TORTOISE

### LINKED ISSUE #71: ENH: Dealing with gradient non-linearities
For gradient systems with significant non-linearities (e.g. connectome scanner, some 7T head-only systems), it is important to account for these as they not only lead to spatial distortion, but also spatially-dependent changes in b-vector encoding. A recent paper quantifies this nicely: https://www.sciencedirect.com/science/article/pii/S1053811919307189
It is described in the HCP pre-processing pipeline as well: https://github.com/Washington-University/HCPpipelines/wiki/FAQ

There are a number of different features that could be related to this:
1. Allowing the user to input a gradient table that describes the non-linearities (since the tables are proprietary); the app could use to this generate the warp, concatenate with any other relevant transforms (e.g. to reference T1 or template space), and generate a gradient deviations file (grad_dev nifti) that encodes the Jacobian of the warp.
2. Have the user supply already unwarped data, along with grad_dev files (e.g. public HCP data) instead of performing the unwarping inside the app
3. Incorporating the gradient deviation files (grad_dev) in the app for tensor fitting and other reconstructions -- now I am not sure whether the existing tools underlying qsiprep currently support the use of grad_dev files (e.g. FSL and MDT are two other software packages that do).

I am hopeful to hear that there would be interest from existing qsiprep users in incorporating these features, as I admit they are not required for the majority of datasets. Our in-house pipeline/BIDS-app (https://github.com/khanlab/prepdwi) actually implements these general steps,  however that pipeline was not designed for longevity and I really would like to shift over to qsiprep since it is so much more feature-rich and is actually well-engineered!!  Happy to help out in with this if I can -- I don't have alot of time to code, and still a newbie in nipype, but there are some other developers here that I could engage if needed.

One potential issue might be that the underlying fitting/reconstruction in Dipy may not yet support gradient deviations. We are currently using FSL (dtifit & bedpost) and MDT (https://github.com/cbclab/MDT) for that since they support using gradient deviation files for fitting. 

PS - great job on putting qsiprep together, really keen to see it become the standard dMRI processing BIDS App in the community, excellent start!

tsalo:
Closes #71.

## Summary

Adds gradient nonlinearity distortion correction to QSIPrep, using TORTOISE V4's standalone tools.

Gradient coils deviate from their nominal linear field. That does two separable things: it displaces voxels, increasingly so away from isocenter, and it means the diffusion gradient actually applied at a voxel is not the one recorded in the
bval/bvec table. This PR corrects both — a spatial gradwarp displacement field for the first, and a voxelwise gradient deviation tensor for the second.

## Interface

```
--gradient-file PATH     scanner gradient coefficients, or an ITK displacement field
--force gradients        apply full 3D correction regardless of ImageType
--ignore gradients       disable gradient correction entirely
```

`--gradient-file` accepts `.grad` (Siemens), `.dat` (GE), `.gc` (TORTOISE binary), and `.nii`/`.nii.gz` (a ready-made ITK displacement field). One file applies to the whole dataset; process multi-site data one site at a time.

How much *spatial* correction a run gets is read from its `ImageType`:

| `ImageType` tag | Spatial correction |
| --- | --- |
| no `DIS` tag | full 3D gradwarp |
| `DIS2D` | through-plane residual only (scanner already did in-plane) |
| `DIS3D` | none (scanner already did all of it) |

`--force gradients` overrides this for data whose tags are absent or untrustworthy. Runs in one correction unit that disagree resolve to the least-correcting value, with a warning, since they are concatenated before HMC and share one field.

## Where it sits in the pipeline

The field is generated once per DWI run from the raw grid and folded into the composed transform, in the order:

```
head motion → gradwarp → susceptibility → (intramodal →) coregistration → (template)
```

matching TORTOISE's `FINALDATA` composite order. It is never applied to the output series as a resampling of its own.

Two geometry rules follow from that ordering, and both are implemented per SDC node rather than per backend:

- **SDC estimation inputs are gradwarp-corrected** wherever the resulting SDC warp is applied *downstream* of gradwarp — DRBUDDI, GRE fieldmaps, SyN. The exception is `eddy`+TOPUP: `eddy` consumes TOPUP's field via `--field` and resamples the raw data itself, baking it in upstream of `ComposeTransforms`, so estimating that field on corrected images would place it in a space it was never measured in. `--pepolar-method DRBUDDI+TOPUP` runs both and corrects only DRBUDDI's inputs.
- **The b0→T1w coregistration reference is gradwarp-corrected** on every backend and every fieldmap branch, since the coregistration affine is applied after gradwarp. There is no TOPUP exception here — that reference is derived from `eddy`'s output, downstream of everything `eddy` applied.

Head motion and eddy current estimation never see the field; they run on the native grid, matching both TORTOISE and HCPPipelines.

## Outputs

`*_space-ACPC_graddev.nii.gz` — the HCP-style voxelwise 3×3 gradient deviation (L) matrix, row-major, as 9 volumes. Written whenever `--gradient-file` is given, **including for `DIS3D` runs**: no scanner can correct the diffusion encoding, because the bval/bvec table holds one value per volume and has nowhere to put spatially varying information. The effective gradient at a voxel is `L @ g`, and because `L` carries scaling and shear rather than pure rotation, both the b-vector and the b-value deviate per voxel.

`GradientWarpDimensions` is also written into the main `desc-preproc_dwi.json`, and the HTML report gains a resolved-plan line beside `HMC Model`.

## Notes for reviewers

**Why standalone tools rather than `--grad_nonlin` passthrough.** Wrapping `CreateNonlinearityDisplacementMap` and `CreateGradientNonlinearityBMatrix` keeps QSIPrep's own final resampling and avoids reimplementing the spherical-harmonic coefficient model. Three upstream hazards drove the wrapper shape and are documented at `qsiprep/interfaces/gradunwarp.py`: the coefficient file is the *first* positional argument (a stale unbuilt duplicate in the TORTOISE tree has them reversed); the built tool reads its `is_GE` argument with `atoi`; and the returned field is the one TORTOISE names `gradwarp_field_inv` and must **not** be inverted here.

**GE coefficient files are refused.** TORTOISE applies a z-origin shift to the field after expanding GE coefficients, and that shift lives in `TORTOISE.cxx` rather than in the standalone binary QSIPrep calls — so we would place the field
at a different z than TORTOISE does. Under the obvious reading of the recentering the sign looks inverted, so reproducing it blind is not safe either. QSIPrep raises rather than applying a field it cannot place. A ready-made ITK field is unaffected (nothing is expanded), as is `graddev`, whose tool does its own self-contained GE recentring. Lifting this needs an empirical check against a TORTOISE run on GE data.

**Known limitations** are enumerated in `docs/preprocessing.rst` under "Gradient nonlinearity correction". Briefly: no `graddev` under `--distortion-group-merge`; the `graddev` map is oriented by a rigid transform TORTOISE re-derives internally rather than by QSIPrep's own coregistration affine (recorded in the sidecar); and DIFFPREP's fieldmap-less `T2Wreg` path estimates its SDC field pre-gradwarp — which turns out to match stock TORTOISE, whose own `EPIREG` gradwarp step cannot load its file. An issue draft for that upstream defect is at `docs/superpowers/upstream/`.


docs/superpowers/upstream/2026-08-27-tortoise-epireg-gradnonlin-bug.md
tsalo, line 1: @mattcieslak this is a bug in TORTOISE flagged by Claude. We should open an issue once Okan's back from vacation.

tsalo:
My attempt on CUBIC is failing on synthstrip, even with `--cleanenv`.

mattcieslak:
The torch environment is messed up in current unstable. I have this fixed in #1111

docs/superpowers/upstream/2026-08-27-tortoise-epireg-gradnonlin-bug.md
mattcieslak: Yes, definitely. In the meantime we can fix it in our TORTOISE fork here github.com/PennLINC/qsiprep_TORTOISE

tsalo:
Thx! I'll hold off on testing until you've merged your PR then.

tsalo:
Testing again now that the other PR has been merged.

docs/superpowers/upstream/2026-08-27-tortoise-epireg-gradnonlin-bug.md
tsalo: I've opened https://github.com/PennLINC/qsiprep_TORTOISE/issues/1.

tsalo:
It is working! The change is so subtle you can't really see it in the reportlet, but if you zoom in and look at the edge of the field of view (way outside the brain), you can see some movement.

mattcieslak (approved):

PR #1110, opened 2026-09-02
Make MRtrix development version available with `--mrtrix-version`

### LINKED ISSUE #1108: Pass complex-valued data to mrdegibbs
## Summary

`mrdegibbs` in the MRtrix3 development branch properly leverages complex-valued data, so we can take the denoised data from the dwidenoise/dwidenoise2 step and pass it to `mrdegibbs` directly instead of only passing along the magnitude component.

We'll also need to use the MRtrix3 development branch, which might require a little effort to figure out.

tsalo:
Closes #1108.

## What this does

`mrdegibbs` is built on the Fourier shift theorem, so it can unring complex-valued data
directly — but only on MRtrix3's development branch. The released `mrdegibbs` reads
magnitude data only. QSIPrep therefore had to throw the phase data away immediately
after denoising, even when it had just been used.

This PR carries complex-valued data through unringing, and puts that behind a new
`--mrtrix-version` flag so nobody is moved onto unreleased MRtrix3 without asking. The
container now ships **both** MRtrix3 installations side by side, and the flag chooses
between them at runtime.

```
--mrtrix-version {stable,dev}    default: stable
```

- `stable` — a released MRtrix3 (currently 3.0.4). Unchanged behavior.
- `dev` — the development branch (currently `3.0.8-2071-gb98b54e9`). Required for
  complex-valued unringing.

## What changed

**Version selection.** `config.workflow.init()` puts the selected installation's `bin/`
first on `PATH`, keeping the other reachable behind it. `PATH` ordering is the only
mechanism that works here: `dwibiascorrect` is a Python script that resolves `mrcalc`,
`dwiextract`, `mrmath`, `mrconvert` and `N4BiasFieldCorrection` through its own `PATH`
lookup, so pinning absolute paths on the nipype interface would still leave the script's
own children resolving against whatever `PATH` says. The baked image order matches
`stable`, so the CLI default and the image default agree and `init()` only has real work
to do for `dev`.

**Complex unringing.** The denoising chain used to track state by buffer position, which
made "is this image still complex?" a question about where you were in the chain. It now
tracks the data domain explicitly (`_ImageChain`), so the magnitude split is inserted at
the last step that can still consume complex data rather than always landing right after
denoising. Selecting `--unringing-method mrdegibbs` with phase data under
`--mrtrix-version stable` logs a warning and unrings the magnitude data, as before.

**ANTs option spelling.** MRtrix3 renamed `dwibiascorrect`'s `-ants.b` / `-ants.c` /
`-ants.s` to `-ants_b` / `-ants_c` / `-ants_s` on the development branch. `DWIBiasCorrect`
now emits the right spelling for the selected version via `_format_arg`. This is a silent
run-killer if you get it wrong, so there is a test that asks the real binaries.

**Report warning.** Runs using `--mrtrix-version dev` get a Bootstrap warning callout at
the top of the subject summary, and the About section names the installation and version
that actually ran — including when no installation was declared, where it says so rather
than claiming a version it never resolved.

**Container.** `Dockerfile.base` copies both trees to `/opt/mrtrix3-stable` and
`/opt/mrtrix3-dev` and declares `MRTRIX3_STABLE_HOME` / `MRTRIX3_DEV_HOME` /
`MRTRIX3_STABLE_VERSION` / `MRTRIX3_DEV_VERSION`. Neither tree contributes to a global
`LD_LIBRARY_PATH` — the development build is now RPATH-relative, so each tree finds its
own libraries and cannot shadow the other. The consuming `Dockerfile` asserts at build
time which tree each command resolves to and that the baked version strings match
`mrinfo -version`, so a stale pin fails the build instead of shipping quietly.

**Unrelated bug fix, included because it blocks the container tests.** `bvec_to_rasb`
raised on *any* `mrinfo` output to stderr. Recent MRtrix3 prints "axes realigned to
approximate RAS" for every non-RAS image, which is every LPS+ image QSIPrep produces. It
now fails on a non-zero return code and logs advisories at debug level.

## Compatibility

- Default behavior is unchanged. `--mrtrix-version` defaults to `stable`, so
  complex-valued unringing is opt-in rather than something existing users are moved onto.
- `--denoise-method dwidenoise2` is unaffected by this flag. `dwidenoise2` exists only on
  the development branch, so it always resolves there regardless of `--mrtrix-version`,
  and `--mrtrix-version stable --denoise-method dwidenoise2` remains a valid mixed-version
  configuration that behaves exactly as it does today.
- Complex-valued *denoising* also already worked under `stable` — 3.0.x `dwidenoise`
  accepts complex input. This flag gates only what happens after denoising.

## Testing

- **Config `init()`** — parameterized over both versions against two `tmp_path` roots:
  selected tree first, other second, no-op when neither variable is set, `RuntimeError`
  when the selected root is declared but its `bin/` is missing, and no `PATH` duplication
  across repeated reloads.
- **`DWIBiasCorrect.cmdline`** — the dot spelling under `stable`, the underscore spelling
  under `dev`.
- **Real binaries, in-container** — asks each `dwibiascorrect` whether it accepts its own
  spelling *and* rejects the other one, with a bogus-option control. The control matters:
  `-help` short-circuits MRtrix3's argument parsing, so under `-help` even
  `-ants.TOTALLYBOGUS` "passes". These skip outside the container.
- **Workflow shape** — the complex/magnitude graph tests parameterized over both versions,
  including the warning emitted at the real log level rather than a lowered one.
- **Report** — the warning appears for `dev`, is absent for `stable`, and the About
  section renders correctly for all four provenance shapes.
- **End to end** — `mrdegibbs` runs on complex data in the container tests.

## Note for reviewers

The branch also carries the design and planning documents under `docs/superpowers/`
(~3,500 lines). They are excluded from the Sphinx build and are safe to skip when
reviewing; say the word and I will strip them from the branch.


docs/superpowers/plans/2026-09-01-complex-mrdegibbs.md
mattcieslak, line 1: I don't think we should commit these plans to the repo. Can you delete before we merge?

qsiprep/interfaces/mrtrix.py
```diff
@@ -444,9 +445,19 @@ class DWIBiasCorrectInputSpec(MRTrix3BaseInputSpec, SeriesPreprocReportInputSpec
         position=-1,
         desc='the output bias corrected DWI image',
     )
-    ants_b = traits.Str(default_value='[150,3]', argstr='-ants.b %s', usedefault=True)
-    ants_c = traits.Str(default_value='[200x200,1e-6]', argstr='-ants.c %s', usedefault=True)
-    ants_s = traits.Str(default_value='4', argstr='-ants.s %s')
+    # The argstr is a bare placeholder: DWIBiasCorrect._format_arg builds the real
+    # flag, because 3.0.x spells these -ants.b and the development branch -ants_b.
+    # The placeholder is load-bearing — nipype formats only traits that declare an
+    # argstr, so removing it would silently emit nothing at all.
```
mattcieslak: I think we should leave these story-telling comments out. This is standard nipype stuff

qsiprep/interfaces/mrtrix.py
```diff
@@ -542,6 +570,12 @@ def _generate_report(self):
 
         input_dwi, denoised_nii, _ = self._get_plotting_images()
 
+        # mrdegibbs emits complex data when it is given complex data, and the report
+        # always shows magnitude. This mirrors SeriesPreprocReport._generate_report,
+        # which this method overrides.
```
mattcieslak: unclear what this comment is referring to

qsiprep/tests/test_interfaces_mrtrix.py
```diff
@@ -109,3 +111,244 @@ def test_dwidenoise2_formats_fslgrad(tmp_path):
...
+    assert '-dimensionality 3' in mrtrix.MRDeGibbs(in_file=in_file, dimensionality=3).cmdline
+
+
+def test_mrdegibbs_report_handles_complex_input(monkeypatch, tmp_path):
+    """Generate the unringing report from complex-valued data, using the magnitude.
+
+    mrdegibbs on MRtrix3's development branch emits complex data when it is given
+    complex data. nibabel's get_fdata() does not raise on a complex image: it emits a
+    ComplexWarning and silently returns the real component, discarding the imaginary
+    part. Without an explicit magnitude conversion the report would be drawn from that
```
mattcieslak, line 172: This is partly a repeat of the earlier comment, I think this is a much better spot for it

qsiprep/workflows/dwi/merge.py
```diff
@@ -435,6 +471,23 @@ def get_buffernode():
     do_unringing = config.workflow.unringing_method in ('mrdegibbs', 'rpg')
     harmonize_b0s = not config.workflow.no_b0_harmonization
 
+    # Only the dwidenoise variants can denoise complex-valued data. Any other method
+    # ignores the phase data and denoises the magnitude data alone.
+    denoise_complex = do_denoise and denoise_method.startswith('dwidenoise') and use_phase
+    # mrdegibbs is built on the Fourier shift theorem and reads and writes complex data
+    # on MRtrix3's development branch, so complex data stay complex through unringing.
+    # The released mrdegibbs cannot, and TORTOISE's rpg is magnitude-only.
```
mattcieslak: TMI

qsiprep/workflows/dwi/merge.py
```diff
@@ -629,15 +683,12 @@ def get_buffernode():
...
-        # Add buffernode for unringed DWI
-        buffernodes.append(get_buffernode())
-
         workflow.connect([
-            (buffernodes[-2], degibbser, [('dwi_file', 'in_file')]),
             (degibbser, ds_report_unringing, [('out_report', 'in_file')]),
-            (degibbser, buffernodes[-1], [('out_file', 'dwi_file')]),
             (degibbser, merge_confounds, [('nmse_text', f'in{step_num}')]),
         ])  # fmt:skip
+        chain.feed(degibbser, 'in_file')
```
mattcieslak, line 687: this is pretty cool. I've never seen this kind of thing in nipype before

Dockerfile
```diff
@@ -1,44 +1,4 @@
...
-    cp /src/dwidenoise2/cpp/cmd/dwi2noise.cpp cpp/cmd/dwi2noise.cpp && \
-    cp -r /src/dwidenoise2/cpp/core/denoise cpp/core/denoise && \
-    cp -r /src/dwidenoise2/share/dwidenoise2/. share/mrtrix3/ && \
-    cmake -B build -GNinja \
-          -DMRTRIX_BUILD_GUI=OFF \
-          -DMRTRIX_ENABLE_GPU=OFF \
-          -DCMAKE_COMPILE_WARNING_AS_ERROR=ON \
-          --preset=release && \
-    cmake --build build --target dwidenoise2 dwi2noise copy-share-data
+ARG BASE_IMAGE=pennlinc/qsiprep-base:20260903
```
mattcieslak: I might have accidentally done this on github. I think we definitely should not be compiling mrtrix here. It should be COPYed in like the rest of the binaries

Dockerfile
```diff
@@ -77,26 +37,22 @@ FROM ${BASE_IMAGE} AS base
...
-    test -d /opt/dwidenoise2/share/mrtrix3/dwidenoise2
+# Pin which tree each command resolves to. The default PATH order matches
+# --mrtrix-version stable; config.workflow.init() reorders it for --mrtrix-version dev.
+# dwidenoise2 exists only in the development tree, so it must fall through to it.
+# Only compiled binaries may be executed here. MRtrix3's dwibiascorrect and the rest of
+# its scripts start with `#!/usr/bin/env python`, and no interpreter exists in this stage
+# -- Python only arrives with the pixi environment layered on in the stages below. Both
+# trees' dwibiascorrect are run for real, against their own -ants.b/-ants_b spelling, by
+# test_dwibiascorrect_options_are_accepted_by_the_real_binary in the test image.
+RUN test "$(command -v mrdegibbs)"      = "/opt/mrtrix3-stable/bin/mrdegibbs" && \
```
mattcieslak, line 45: is it common to test things in docker builds? I like it

mattcieslak (commented):
Looks good so far, a couple things worth discussing before we merge

mattcieslak (commented):
Looks good overall. A couple points worth discussing before we merge though

qsiprep/interfaces/mrtrix.py
tsalo: Basically, without an override, `_generate_report` would load the complex-valued data with `img.get_fdata()`, which only returns the _real_ component (not the complex data or the magnitude component, as desired).

Dockerfile
tsalo: It's caught stuff that wasn't on the path for us before. Probably worth dropping the checks every once in a while when things are stable, but it doesn't cost much time in the build so I find it very useful.

PR #1113, opened 2026-09-09
Stop DWI summary reportlet duplication

tsalo:
With ``--anat-modality none`` there is no anatomical image, so the subject summary is written with a ``dwi`` suffix and shows up at the top of the Diffusion section rather than at the top of the report.


tsalo:
I verified it worked in the `drbuddi_rpe` CI reports.

PR #1115, opened 2026-09-09
Add white matter boundary to coregistration figure

### LINKED ISSUE #1018: HTML report - enhanced view of the DWI to T1w coregistration
## Summary
Hi, I would like to suggest a better way of depicting the alignment of the DWI and T1w in the HTML report. Instead of switching between the T1w and DWI, I would suggest to take the pial surface (GM/CSF boundary) from the T1 segmentation and overlay (without animation) onto the DWI.

## Additional details
It is kinda hard to see if the images are really aligned, since image contrast is so much different between the T1 and DWI (plus it often seems at first glance that they are not aligned properly, when in fact they are).

tsalo:
Closes #1018. This doesn't make it a static figure, but it does add a white matter boundary line to match fMRIPrep and make it easier to interpret the plot. I also remodeled some of the reportlet generation interfaces to match fMRIPrep.

- Pass white matter segmentation to coregistration reportlet.
- Replace ANTSRegistrationRPT with ants.Registration + SimpleBeforeAfterRPT.
- Replace niworkflows SimpleBeforeAfterRPT with nireports.


qsiprep/workflows/dwi/registration.py
```diff
@@ -205,10 +227,12 @@ def init_direct_b0_acpc_wf(write_report=True, name='b0_anat_coreg'):
 
     workflow = Workflow(name=name)
 
-    # Defines a coregistration operation
+    # Defines a coregistration operation. The reportlet for this workflow is
+    # ACPCReport's translation-vs-rigid comparison, not a registration flicker,
+    # so this node renders nothing.
     ants_settings = str(load_data('intermodal_ACPC.json'))
```
mattcieslak: what does it mean that this renders nothing?

mattcieslak (commented):
would this also fix the zooming behavior in some of the masked svgs?

tsalo:
I don't think so. We were seeing that zooming in plots with the WM boundary lines. I suppose nireports might have improvements in their SimpleBeforeAfterRPTs compared to niworkflows though.

qsiprep/workflows/dwi/registration.py
tsalo: I think the report generated by the ANTSRegistrationRPT here was not being passed to `outputnode` to be written out with a DerivativesDataSink, so it ended up just sitting in the working directory. The ACPCReport was the one being written out to the output directory, which is a behavior that is retained in this PR.

tsalo: I'll just remove the comment though- it's not helpful

tsalo:
I just confirmed that the scaling issue is still present in `desc-sdcdrbuddi_dwi.svg` in diffprep_csdsi_rpe_series at least, though my PR doesn't touch that figure.

PR #1116, opened 2026-09-09
Read the part-phase companions from qsiplan

mattcieslak:
qsiplan now models complex-valued DWI: a part-phase image is the companion of its magnitude, carried in the plan as PreprocUnit.dwi_phase_files keyed by the magnitude path. This PR has qsiprep consume that directly instead of globbing.

- collect_data no longer pre-filters on part, so every part reaches qsiplan, which performs the authoritative split (real/imag are dropped with a warning).
- init_merge_and_denoise_wf reads the phase companion from unit.dwi_phase_files rather than re-globbing the layout; the downstream complex-denoise plumbing is unchanged, so --dwi-phase-correction composes on top of it.
- Require qsiplan >= 0.4.0, and add an end-to-end contract test that a complex acquisition reaches the plan as a magnitude with a phase companion, with no phase leaking into any series list or sidecar override.
- Docs: real/imaginary parts are ignored with a warning, not an error.


tsalo (approved):
The changes look reasonable, but the idea of grabbing all files (both mag and phase) in the first data collection step makes me wonder how this would work with other file collections, like multi-echo DWI. Normally I'd collect the representative file (echo-1, part-mag) and have a second collection function to get the associated files for it.

Of course, we don't support multi-echo DWI (yet), so it's not an issue at the moment.

PR #1118, opened 2026-09-09
Use ants Affine Initializer (antsAI) for coregistration

mattcieslak:
We were running into issues in HBCD where the infants would be significantly rotated between the anatomical scans and the dMRIs. Other times, the infant was rotated to the point of almost being non-supine for the entire scan, which was causing spatial normalization to fail.

This PR inserts antsAI any time before a registration occurs. This includes

 * Anat to dwi
 * Bonus T2w to dwi
 * Intramodal template

## Update

It turned out that we had unknowingly updated ANTs to >2.6.0 by switching its installation to pixi from qsiprep_build. This ultimately benefited us for antsAI, but it also revealed that FSL was being installed via pixi too. This PR adds a script that forces installed FSL components to have versions consistent with a tagged FSL release. As some further cleanup we also reduced the FSL presence in qsiprep by replacing fslmaths (fieldmap workflows) and fsl.Split/fsl.Merge (eddy path) with nibabel/scipy interfaces. Header/orientation/TR fidelity verified against fslmaths so the eddy workflow no longer depends on fsl-avwutils.


cookpa:
This needs ANTs >= 2.6.0 to avoid a nasty ITK bug that messes up the multi-start optimizer.

Also, you might consider using a slightly smaller arc fraction, eg 0.45, so that the 20-degree steps hit closer to 0, eg 0.45 = 81 degrees so you'll get -81, -61, -41, ... -1, 21,... whereas with arc = 0.5 you'll get -90, -70, ... -10, 10, 30,...

It's probably enough to capture but I always like to have a start point close to 0 as many studies make some effort to align head position.

PR #1121, opened 2026-09-10
Raise an exception if no files are found for sessions in `--session-id`

### LINKED ISSUE #1120: All sessions silently collected when no DWI available for `--session-id`
## Summary

I noticed this when trying to run QSIPrep with `--session-id 1` in a dataset with 01 and 02 sessions. Instead of selected session 01 or raising an exception, it collected data from both 01 and 02.

## Additional details

- QSIPrep version: unstable (pre-26.2.0)
- Docker version:
- Singularity version:

tsalo:
Closes #1120.

- Instead of wiping the session filter when no DWI files (or the associated anatomical modality for `--anat-only` mode) are present in the `--session-id` sessions, raise an error.


mattcieslak:
I think we may want to deprecate anat-only. smriprep is very good, and qsiplan is totally dwi-centric.

tsalo:
I'm all for it. I think the only thing we need for that is a way to ingress sMRIPrep derivatives, which will involve calculating an sMRIPrep T1w-to-ACPC transform.

PR #1144, opened 2026-09-17
Rename CLI parameters

### LINKED ISSUE #1089: Deprecate `--b1-biascorrect-stage` parameter and remove in 27.0.0
We can instead have `--dmri-biascorrect <auto|n4|none>` to match `--anat-biascorrect`.

If anyone needs the legacy approach, they can use an older version of QSIPrep.

When using auto, there should be a warning box in the HTML report since we don't know how well the check (using ImageType) generalizes.

### LINKED ISSUE #1114: Toggle intramodal template generation with new `--dwiref-definition` parameter
## Summary
The idea is to align with fMRIPrep's new `--bold-coreg-level <run|session|subject>` flag, although we may want a `distortion-group` value instead of `run`.

The default for `--dwi-coreg-level` would be the lowest level (distortion-group/run) to match current behavior. The default for `--intramodal-template-iters` could be set to our recommended value when applying the intramodal template (3 I think?), though it would have no effect for distortion-group/run.

### LINKED ISSUE #1122: Rename `--b0-to-anat-transform` to `--dwi2anat-dof`
## Summary

This would better match fMRIPrep and ASLPrep.

- `--b0-to-anat-transform Rigid` --> `--dwi2anat-dof 6`
- `--b0-to-anat-transform Affine` --> `--dwi2anat-dof 12`
- `--dwi2anat-dof 9` would be a new option

### LINKED ISSUE #1129: Rename `--intramodal-template-transform` to `--dwiref-construction-transform`
Related to #1114, #1122, and #1128.

`--intramodal-template-transform <Rigid|Affine|SyN|BsplineSyN>` --> `--dwiref-construction-transform <Rigid|Affine|SyN|BsplineSyN>`

tsalo:
Closes #1114. Closes #1129. Closes #1122. Closes #1089.

Four user-facing options are renamed to match fMRIPrep and ASLPrep conventions, the
pre-0.17 "legacy" DWI bias-correction path is removed, and the dwiref derivatives are
reorganized so every coregistration-reference level has a file on disk.

**This PR must not merge before PennLINC/qsirecon#404**, which teaches QSIRecon to read
the renamed reference image. Merging this first breaks QSIRecon ingestion.

## Breaking changes

Every removed option is gone outright, with no deprecation period — an old command line
fails with argparse's "unrecognized argument" rather than running with different meaning.

| Removed | Replacement |
|---|---|
| `--b1-biascorrect-stage {final,none,legacy}` | `--dwi-biascorrect {n4,auto,none}` |
| `--dwi-no-biascorr` | `--dwi-biascorrect none` |
| `--b0-to-anat-transform {Rigid,Affine}` | `--dwi2anat-dof {6,12}` |
| `--b0-to-t1w-transform` (already deprecated) | `--dwi2anat-dof {6,12}` |
| `--intramodal-template-transform` | `--dwiref-construction-transform` |
| `--intramodal-template-iters` | `--dwiref-construction-iters` |

`--dwiref-construction-iters` now defaults to 2 and rejects values below it. It no longer
toggles the feature — that moved to `--dwiref-definition`. Only the linear template branch
clamped the iteration count; the default nonlinear branch handed it straight to
`antsMultivariateTemplateConstruction2`, so an unvalidated `0` would have reached it.

## `--dwi-biascorrect`

`n4` (default) and `none` behave as `--b1-biascorrect-stage final` and `none` did.
`legacy` has no replacement: it applied N4 before concatenation, and anyone who needs it
can use an older QSIPrep.

`auto` skips N4 when every DWI feeding an output is flagged `NORM` in its BIDS
`ImageType` metadata. It is deliberately conservative — a mixed set is corrected, and
missing or unreadable metadata counts as un-normalized, since running N4 unnecessarily is
milder than skipping it when it was needed.

The decision is made **once per final output**, over the union of its constituent
correction units. Deciding per unit would let `--distortion-group-merge` concatenate a
corrected series with an uncorrected one.

Because `auto` is a heuristic whose generalization across vendors isn't established, the
DWI report carries a warning box when it made the call, and states the resolved outcome
(`auto (applied)` / `auto (skipped)`) rather than just the mode.

## `--dwi2anat-dof`

`6` maps to a rigid ANTs transform, `12` to affine. **9 is not offered**:
`antsRegistration` has only Rigid (6), Similarity (7) and Affine (12), unlike the
FLIRT/`mri_coreg` path fMRIPrep reaches 9 through. Adding it needs a second registration
backend — filed separately.

## `--dwiref-definition`

New option, `{distortion-group,subject}`, default `distortion-group`. It replaces "a
nonzero iteration count" as the toggle for what was the intramodal template. `session` is
deferred (the remainder of #1114).

Derivative naming now separates two orthogonal facts: **`space` carries a dwiref's
level**, and **`desc-coreg` marks its role** — the one whose transform final resampling
actually uses. That role moves: with a subject-level dwiref the template is the
coregistration target, because `ComposeTransforms` prefers the intramodal affine when both
exist.

All dwiref products move from `anat/` to `dwi/`:

| Before | After |
|---|---|
| — | `dwi/…_dwiref.nii.gz` (new, every unit; `_desc-coreg_` at distortion-group level) |
| `anat/sub-X_space-ACPC_desc-intramodal_dwiref.nii.gz` | `dwi/sub-X_space-subject_desc-coreg_dwiref.nii.gz` |
| — | `dwi/sub-X_space-ACPC_dwiref.nii.gz` (the same template, resampled) |
| `anat/sub-X_from-intramodal_to-ACPC_…xfm.mat` | `dwi/sub-X_from-subject_to-ACPC_…xfm.mat` |
| `anat/sub-X_space-ACPC_desc-agreement_dwiref.nii.gz` | `dwi/sub-X_space-subject_desc-agreement_dwiref.nii.gz` |
| `dwi/…_space-ACPC_dwiref.nii.gz` | `dwi/…_space-ACPC_desc-preproc_dwiref.nii.gz` |

Three corrections fall out of this:

- The **midpoint-space template is now shipped**. Previously only the ACPC-resampled image
  was written, while `from-intramodal_to-ACPC` mapped out of the midpoint space — so the
  transform applied to no file in the output.
- The **agreement map's space tag was wrong**. It is computed from midpoint-space aligned
  images and was labelled `space-ACPC`.
- The **post-resampling reference gains `desc-preproc`**, matching the series it
  accompanies. Without it, an output group with no distinguishing entity renders to the
  same path as the subject template.

`templateQC`, the agreement map, and the per-group `from-orig_to-*` transforms remain
linear-only — `antsMultivariateTemplateConstruction2` exposes no per-input aligned images
— so with the default `BSplineSyN` none of the three is written. Unchanged, but easy to
miss.

## Collision guard

`init_single_subject_wf` now refuses to build a subject whose output names would render to
the same BIDS path, rather than silently overwriting.

QSIPlan uniquifies same-named correction units with a `+N` suffix (`sub-01`, `sub-01+2`),
an in-memory key that is not a BIDS entity. Four DWI runs with the phase-encoding
direction only in the sidecar and two different `TotalReadoutTime` values are enough to
trigger it — an ordinary acquisition. Reported upstream; this guard protects users until
it's fixed.

The check parses what `get_source_file` actually builds rather than a synthetic path. The
subject and session entity patterns are path-anchored, so a probe directory makes the
`ses-1`/`ses-1+2` case look collision-free under pybids >= 0.19 while the real path still
collides. Verified against 0.15.6 and 0.19.0.


tsalo:
@copilot resolve the merge conflicts in this pull request

qsiprep/cli/parser.py
```diff
@@ -467,6 +467,12 @@ def _bids_filter(value, parser):
             '--infant is given.'
         ),
     )
+    g_anat.add_argument(
```
tsalo: This should have been removed.

qsiprep/utils/sentry.py
tsalo: This should have been removed.

tsalo:
Well, Copilot did a bad job. I'll avoid it from now on.

docs/preprocessing.rst
```diff
@@ -366,7 +374,13 @@ Transforms
 
 .. important::
 
-  *QSIPrep* does not currently write out the coregistration transform from dwiref space to ACPC space.
+  With ``--dwiref-definition subject`` *QSIPrep* builds a single midpoint dwiref for
+  the subject, registers it to the anatomical once, and has every DWI group inherit
+  that transform. It writes the template in its own space
+  (``sub-<label>_space-subject_desc-coreg_dwiref.nii.gz``), the same template
+  resampled into ACPC (``sub-<label>_space-ACPC_dwiref.nii.gz``), and the transform
+  between them (``sub-<label>_from-subject_to-ACPC_mode-image_xfm.mat``), so that
```
mattcieslak: is this the convention fmriprep uses too?

qsiprep/cli/parser.py
```diff
@@ -54,9 +46,7 @@ def _build_parser(**kwargs):
     from pathlib import Path
 
     # Deprecated options: {option string: (version it is removed in, what happens instead)}
-    deprecations = {
-        '--b0-to-t1w-transform': ('27.0.0', 'Please use `--b0-to-anat-transform` instead.'),
-    }
+    deprecations = {}
```
mattcieslak, line 49: would it make sense to leave this infrastructure in for future deprecations?

qsiprep/cli/parser.py
```diff
@@ -525,10 +490,11 @@ def _bids_filter(value, parser):
...
         help=(
-            'Whether to run N4 bias field correction on the anatomical images. This is '
-            'separate from --b1-biascorrect-stage, which governs the DWIs only. '
-            '"n4" always runs it; scanner-side intensity normalization (Siemens NORM, '
-            'for example) does not remove the need for it. '
+            'Whether to run N4 bias field correction on ANATOMICAL images. '
+            'Note this is separate from --dwi-biascorrect, which only governs '
+            'the DWIs. '
+            '"n4" (default) always runs it; scanner-side intensity normalization '
+            '(e.g. Siemens NORM) does not remove the need for it. '
```
mattcieslak:
```suggestion
            '(e.g. Siemens NORM) does not necessarily remove the need for it. '
```

qsiprep/workflows/dwi/pre_hmc.py
```diff
@@ -26,6 +26,7 @@ def init_dwi_pre_hmc_wf(
     unit,
     orientation,
     source_file,
+    do_biascorr=True,
```
mattcieslak: minor nitpick: isn't having this default to True the legacy behavior that this PR is getting rid of?

qsiprep/workflows/base.py
```diff
@@ -468,16 +518,20 @@ def init_single_subject_wf(subject_id: str, session_ids: list):
         for unit in preproc_units
     }
 
+    # The resolved dwiref level for this subject, which may fall back below the
+    # requested one. Everything downstream keys off the resolved value: the
+    # derivative labels must describe what was actually built.
     make_intramodal_template = False
```
mattcieslak: this comment is totally opaque to me

qsiprep/workflows/base.py
```diff
@@ -512,13 +566,22 @@ def init_single_subject_wf(subject_id: str, session_ids: list):
...
         ])  # fmt:skip
 
+        # The level this dwiref actually is, from the RESOLVED definition, so that
+        # adding `session` later is a data change rather than a code change.
+        dwiref_space = config.workflow.dwiref_definition
+
+        # The template in its own midpoint space. desc-coreg marks it as the image
+        # whose transform final resampling uses -- which at this level it is, since
+        # ComposeTransforms discards the per-unit registration's result.
+        # from-<level>_to-ACPC maps out of exactly this space, so that transform now
```
mattcieslak: So we get a native-to-dwi template transform in the derivatives along with a dwi template to acpc transform?

qsiprep/workflows/base.py
```diff
@@ -616,18 +700,21 @@ def init_single_subject_wf(subject_id: str, session_ids: list):
...
             t2w_sdc=t2w_available_for_sdc(subject_data, selection, config.workflow.anat_modality),
             anatomical_template=anatomical_template,
+            do_biascorr=do_biascorr,
         )
         dwi_finalize_wf = init_dwi_finalize_wf(
             unit=unit,
             name=dwi_preproc_wf.name.replace('dwi_preproc', 'dwi_finalize'),
             output_prefix=naming_name,
             source_file=source_file,
+            do_biascorr=do_biascorr,
```
mattcieslak, line 682: do these two do_biascorrs respect the user-requested stage?

qsiprep/workflows/base.py
```diff
@@ -616,18 +700,21 @@ def init_single_subject_wf(subject_id: str, session_ids: list):
...
         source_file = get_source_file(list(unit.dwi_files), naming_name, suffix='_dwi')
         output_wfname = output_fname.replace('-', '_')
+        do_biascorr = biascorr_by_output[final_output_name]
         dwi_preproc_wf = init_dwi_preproc_wf(
             unit=unit,
             output_prefix=naming_name,
             source_file=source_file,
             t2w_sdc=t2w_available_for_sdc(subject_data, selection, config.workflow.anat_modality),
             anatomical_template=anatomical_template,
+            do_biascorr=do_biascorr,
```
mattcieslak, line 675: still not sure this is needed anymore

docs/preprocessing.rst
tsalo: fMRIPrep has desc-hmc and desc-coreg dwirefs, but the new --bold-coreg-level option writes out the space-subject and/or space-session dwirefs as well.

qsiprep/cli/parser.py
tsalo: I think that, as long as we write a note at the top of this PR, we can always find it again later. Still, I left it in for now if you want me to keep it.

docs/preprocessing.rst
tsalo: It might make sense to put `space-distortiongroup` in the dwiref when it's not "subject". That way there's no difference between dwiref levels.

qsiprep/workflows/dwi/pre_hmc.py
tsalo: Yeah the default's worth getting rid of. I'll do that.

qsiprep/workflows/base.py
tsalo: It should respect the user-requested toggle, but there's no stage anymore.

qsiprep/workflows/base.py
tsalo: I think the new comment is much clearer.

qsiprep/workflows/base.py
tsalo: That's correct. I know it's not really useful because we can't do fit/apply stuff with QSIPrep, but I think it's worth it to hold only the transforms, just in case we figure out a way to do it.

mattcieslak: I think it's still useful bookkeeping! And these are small files

qsiprep/workflows/base.py
tsalo: The one in dwi_finalize_wf is needed. The one for dwi_preproc_wf could be dropped, but it could mess up the boilerplate a bit.

mattcieslak (approved):
I made a couple minor changes, but I think this is ready to go!

tsalo:
Sweet! I'll merge once CI passes.

PR #1147, opened 2026-09-17
Apply Jacobian weighting to all spatial distortion corrections

### LINKED ISSUE #1133: Document state of Jacobian weighting by SDC approach
TOPUP and DRBUDDI do Jacobian weighting, but the current GRE fieldmap, SyN, and T2WReg do not (AFAWK).

tsalo:
Closes #1133.

## Why

Correcting a spatial distortion moves signal between voxels. Where the acquisition compressed a region, that region's signal ended up in fewer voxels at higher intensity. Restoring the correct shape without rescaling leaves those
voxels too bright. The fix is to multiply by the local volume change, which is the Jacobian determinant of the correction.

Three of the eight correction paths below were modulated before this PR. Five were not. Preprocessed intensities were therefore not comparable across `--sdc-method` and `--hmc-method` choices, and the outputs did not record which
corrections had been modulated.

| Correction | Modulated before | Modulated now |
|---|---|---|
| Gradient nonlinearity | no | **yes** (QSIPrep) |
| Susceptibility — TOPUP | yes (inside `eddy`) | yes (inside `eddy`) |
| Susceptibility — DRBUDDI | yes, empirically | **yes, analytically** |
| Susceptibility — GRE fieldmap | no | **yes** (QSIPrep) |
| Susceptibility — SyN | no | **yes** (QSIPrep) |
| Susceptibility — TORTOISE T2Wreg | no | **yes** (QSIPrep) |
| Eddy current — FSL `eddy` | yes (inside `eddy`) | yes (inside `eddy`) |
| Eddy current — TORTOISE DIFFPREP | no | **yes** (QSIPrep) |

## How

`ComposeJacobianWeights` (`qsiprep/interfaces/jacobian.py`) computes the analytic determinant of the composed native-space distortion warps. It keys the computation on unique warp combinations, so a typical run makes one or two ANTs
calls instead of one per volume. `ApplyJacobianWeights`, renamed from `ApplyScalingImages`, resamples those maps onto the output grid through the intramodal and coregistration transforms it already applies, then multiplies
them into the data.

DRBUDDI, GRE, SyN, and T2Wreg all deliver their warps into the same `fieldwarps` input, so one node serves every backend. There is no per-backend code.

Weights are applied to data that has already been resampled once. This does not add a resampling pass.

### Scope

Only native-space distortion corrections contribute: gradient nonlinearity, susceptibility, and eddy current. Head motion, coregistration, the intramodal template warp, and template-space normalization are excluded. Modulating by a
spatial-normalization warp performs VBM-style volume modulation, which would change DWI signal intensities and the models fitted to them.

Excluding head motion also does not affect where the remaining determinants are evaluated. HMC is the outermost transform in the pull-back, so its determinant is a constant factor.

### DRBUDDI

DRBUDDI previously supplied an empirical intensity ratio (`undistorted_reference / blip_*_b0_corrected`). This PR replaces it with the analytic determinant of DRBUDDI's own warps and removes the `sdc_scaling_images` channel from nine files.

This changes DRBUDDI output values. The ratio included interpolation noise and contrast differences that do not belong in a modulation weight. The change is validated below, but reviewers should know that numbers move on the most-used
SDC path.

## New user-facing surface

- **`--jacobian-weighting` / `--no-jacobian-weighting`**, default on. It controls only the modulation QSIPrep applies itself. FSL `eddy` applies its own modulation for eddy-current and TOPUP susceptibility when its resampling method is `jac`, which is the default. That happens inside `eddy`, and this flag does not affect it.
- **`*_space-ACPC_desc-jacobian_dwimap.nii.gz`**, written whenever QSIPrep applied weights. The sidecar gives `JacobianWeightIndex` (zero-based, one entry per DWI volume), `AppliedCorrections`, and `UnmodulatedCorrections` with a reason where coverage is partial.

No file is written when all of the modulation happened inside `eddy`. A unity map is not written in that case, because QSIPrep holds no map at all.

Dividing by the indexed weight volume reverses the multiplication at the point it was applied. It does not recover an unmodulated series, because denoising and bias-field correction run after resampling and do not commute with it.

## Validation

Four questions were settled by measurement, each with a committed test.

**Correction direction.** Total signal is conserved for the correct direction:

| | error vs raw total |
|---|---|
| multiply by \|det\| | 0.27% |
| divide instead | 37× worse |
| no weighting | 27× worse |

**ANTs transform ordering.** Two non-commuting transforms measured 10.0mm, which confirms the first-listed transform is innermost. The head-motion exclusion above depends on this.

**DRBUDDI equivalence.** On real reverse-phase-encode data, DRBUDDI's own `_JAC / _corrected` matches the analytic determinant at r = 0.9996, with 95th percentile relative error around 1.1%, in both blip directions. This is the basis for replacing the ratio images.

**TORTOISE eddy-current reconstruction.** The 24-parameter Okan quadratic transform, reconstructed from DIFFPREP's transformation file, reproduces TORTOISE's `_moteddy.nii` at r = 0.99922 on real data.

## Known limitations

- **`--distortion-group-merge` outputs get no Jacobian derivative.** The data was weighted during resampling. Only the published map is missing, because the merge path builds its derivatives workflow without the jacobian inputs.
- **`--eddy-config` with `"method": "lsr"`** means `eddy` did not Jacobian-modulate, and QSIPrep cannot add it afterward, because `eddy` has already resampled the data. The run warns and the sidecar records the gap.
- **TORTOISE `correction_mode: cubic`** is not supported for the eddy-current Jacobian. That contribution is skipped and the gap is recorded. The run does not abort, and the quadratic formula is not applied to cubic parameters.

## Testing

Local suite: **861 passed, 23 failed, 5 errors**. The failures and errors match the pre-branch baseline, which is caused by a stale editable QSIPlan in the dev environment (`PreprocUnit.dwi_phase_files`). One new test fails locally for that same reason and passes in the container. Passing tests went from 743 to 861.

### Still needs a CI run on the real datasets

- numeric weight sanity on real data (medians, sidecar bounds)
- the DRBUDDI numeric delta against a pre-change run
- confirmation that each integration dataset resolves to the `CorrectionMethod` the marker tests assume

The marker tests check which configurations write a derivative by inspecting the constructed workflow. This caught one wrong fixture expectation, but it does not execute the workflow.

## Two pre-existing bugs found, not fixed here

Both are unrelated to this work and should get their own issues.

- `qsiprep/workflows/fieldmap/base.py:166` calls `init_pepolar_unwarp_wf` with an `omp_nthreads` argument the function does not accept. The PEPOLAR branch is currently unreachable from all three callers of `init_sdc_wf`, so this does not fail today. It will raise `TypeError` if that branch is ever reached.
- `ComposeTransforms` has a `ThreadPoolExecutor` implementation gated on a `num_threads` input that nothing in the repository sets. It runs single-threaded in production.

## Review notes

Two areas worth a closer look.

- The combined gradwarp-and-SDC path is correct, but its only regression fixture uses a spatially uniform eddy-current map. A uniform map is unchanged by transport, so that fixture cannot detect a transport error in this path.
- Brain masks are reconciled by resampling onto a grid rather than by transporting through a warp. Three separate defects during review came from confusing those two operations. The mask handling is verified correct for the DRBUDDI case and is not proven for the general case.


qsiprep/cli/parser.py
```diff
@@ -719,6 +720,20 @@ def _bids_filter(value, parser):
             'https://github.com/PennLINC/qsiprep/blob/main/qsiprep/data/eddy_params.json'
         ),
     )
+    g_hmc.add_argument(
```
tsalo, line 692: @mattcieslak WDYT of rolling this into `--ignore` instead? fMRIPrep has `--ignore fmap-jacobian` and `--force fmap-jacobian`, but since QSIPrep uses Jacobian weighting for all distortion correction I'm thinking `--ignore jacobian`. That way we don't need a new parameter.

tsalo: I made the switch.

PR #1151, opened 2026-09-18
Remove `--anat-only` flag

### LINKED ISSUE #1150: Remove `--anat-only` flag

tsalo:
Closes #1150.

- Remove the `--anat-only` flag. If users want to only process anatomical data, something like sMRIPrep is more fully featured.


qsiprep/tests/test_cli.py
```diff
@@ -683,18 +681,17 @@ def test_multi_t1w(data_dir, output_dir, working_dir):
     anat_dir = Path(writable_bids_dir) / 'sub-PNC' / 'anat'
```
tsalo: One thing I'm not sure about is whether the `multi_t1w` test is still useful with `--anat-only` removed. Is this the only test that covers the intramodal-template stuff?

